Issue #18, opened 2018-06-19
Bundle babel-polyfill


jonyrock:
It's a problem starting without webpack: 

![image](https://user-images.githubusercontent.com/1022757/41589769-0781b280-73be-11e8-9028-afb7f7c0fb0b.png)

jonyrock:
Basically I need a way to run it without node_modules

Issue #11, opened 2018-06-14
deb package 


jonyrock:
I like how Grafana does it: 

https://grafana.com/grafana/download?platform=linux

just installation from source and everything works

jonyrock:
a nice alternative is to make installation as node for whole server

rozetko:
I'll try to use [fpm](https://github.com/jordansissel/fpm). Seems like exactly what we need.

rozetko:
Actually, I would create not only deb package, but also rpm, like Grafana does

jonyrock:
@rozetko Can I ask you only add a note why you close it

rozetko:
We don't need packages already because we have everything bundled and precompiled in .tar.gz archive

PR #19, opened 2018-06-19
Python compilation

rozetko:
Implementation of https://github.com/hastic/hastic-server/issues/11

Hey, @jonyrock

Check it out

After implementing https://github.com/hastic/hastic-server/issues/18 we can make release with _.tar.gz_ archive and have just 1 dependency left (node.js)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ node_modules/
 
 .vscode/
 package-lock.json
-__pycache__/
 *.pyc
 *.txt
 *.log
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -5,7 +5,6 @@ node_modules/
 
 .vscode/
 package-lock.json
-__pycache__/
```
jonyrock: Not sure I understand that if it is a good idea

rozetko: Actually, I just moved it closer to python code (to analytics/ directory)

Issue #24, opened 2018-06-22
Analytics / server messaging 


jonyrock:
We want to use zeromq for messaging between node.js / python

It will allow to separate processes and debug easier

- [x] ZeroMQ Node/Python worker refactoring
- [x] Docs to new installation
- [x] Remove redundant deps after refactoring
- [x] Return analytics status 
- [x] Fix building for production

jonyrock:
https://stackoverflow.com/questions/18499497/how-to-process-sigterm-signal-gracefully

jonyrock:
@rozetko 
I think there could be problems with using pyinstaller with pyzmq  

https://github.com/zeromq/pyzmq/issues/999

Issue #48, opened 2018-07-02
Analytics / Server logs with prefix


jonyrock:
Imagine log which looks like this:

```
[ANALYTICS] something
[ANALYTICS] something more
[SERVER] oh oh
[SERVER] more logs
[ANALYTICS] learning
```

@rozetko

jonyrock:
Also there was comment https://github.com/hastic/hastic-server/pull/49/files/17003e025b86e231b9229a4254dd2209589f34c3#r199461609

to add log level

amper43:
- [ ] look for logging framework that can be useful
add logs about:
- [ ] add logs for different levels
- [ ] for debug level add logs for DB responces, logs about communication with analytics
- [ ] exceptions (not each exceptions logged now)
- [ ] make analytics logs visible in non docker case
- [ ] requests to server
- [ ] for logs like 'OK' add components that reported
- [ ] probably add more logs to grafana-datasource-kit

jonyrock:
@amper43 what is the status of this issue?

amper43:
@jonyrock checkboxes updated. lets
- [ ] add timestamps to server logs
and fix previous points:
- [ ] log requests to server
- [ ] make analytics logs visible in non docker case

jonyrock:
@amper43 not sure I understand meaning of "log requests to server"

amper43:
@jonyrock GET/POST requests that server received

PR #49, opened 2018-07-02
Analytics server messaging #24 v2

jonyrock:


analytics/server.py
```diff
@@ -0,0 +1,63 @@
+import config
+import json
+import logging
+import zmq
+import sys
+
+from worker import Worker
+
+root = logging.getLogger()
+root.setLevel(logging.DEBUG)
```
rozetko: Guess it would be better if we set logging level based on environment (production / development)

jonyrock: @rozetko thank you for reply, I added it as subtask to https://github.com/hastic/hastic-server/issues/48

Issue #66, opened 2018-07-06
Return version of server 


jonyrock:
We can face a problem where user has an older version of hastic-panel and we need to return  version of hastic-server.  Also it is necessary for bug reports. 

So we need to 
- [ ] keep track of current version of server with is the same as release 
- [x] return server version when get status (root) url

The first time we get this problem: https://github.com/hastic/hastic-server/pull/65

jonyrock:
we can get it from `package.json`

rozetko:
Guess we can close it

PR #76, opened 2018-07-14
Return version of server#66

sankerust:
A PR for this issue:
https://github.com/hastic/hastic-server/issues/66

Please advise what to add.
Right now simply logs server version whne server starts, taken from package.json.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hastic-server",
-  "version": "1.0.0",
+  "version": "0.1.4-alpha",
   "description": "REST server for managing data for analytics",
   "scripts": {
     "start": "node dist/server.js",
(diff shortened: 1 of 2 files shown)


rozetko:
You should add it to `/` url which returns status
Response should look like
```json
{"server":"Ok", "analyticsReady":true, "version": "0.2.0-alpha"}
```

sankerust:
We can add a new field in package.json
Something like "MinRequiredPanelVersion" or "CompatiblePanelVersion" and add there compatible panel versions.
And return this in status, same as server version.
Then the panel will be able to read the version from the status and throw an Error, if it doesnt fir the requirement.
What do you think @jonyrock ?

jonyrock:
Lets merge what we have for now

Issue #85, opened 2018-07-18
NeDB integration


rozetko:
To do https://github.com/hastic/hastic-server/issues/58 we should integrate NeDB first

PR #87, opened 2018-07-18
NeDB integration #85

rozetko:
Closes https://github.com/hastic/hastic-server/issues/85
Now we can [insert](https://github.com/louischatriot/nedb#inserting-documents) and [find](https://github.com/louischatriot/nedb#finding-documents) items in corresponding databases


server/src/config.ts
```diff
@@ -10,6 +10,10 @@ export const ANALYTICS_PATH = path.join(__dirname, '../../analytics');
 
 export const DATA_PATH = path.join(__dirname, '../../data');
 
+export const ANALYTIC_UNITS_DATABASE_PATH = path.join(DATA_PATH, 'analyticUnits.db');
```
jonyrock: why do we need 3 dbs?

rozetko: It is like a best practice from [nedb docs](https://github.com/louischatriot/nedb#creatingloading-a-database) to store multiple collections
```javascript
// Of course you can create multiple datastores if you need several
// collections. In this case it's usually a good idea to use autoload for all collections.
db = {};
db.users = new Datastore('path/to/users.db');
db.robots = new Datastore('path/to/robots.db');
```

Seems like not a best way to store everything in one object, doesn't it?

Issue #91, opened 2018-07-23
Missing communication between analytics and server in production


jonyrock:
Based on conversation with @petrk94
 https://github.com/hastic/hastic-server/issues/90#issuecomment-407084240

jonyrock:
@petrk94 we fixed it !

PR #95, opened 2018-07-24
Missing communication between analytics and server in production #91

jonyrock:
Closes https://github.com/hastic/hastic-server/issues/91
* Correct implementation of IPC communication
* Server stopping
* Small bugfixing

### DIFF
--- a/server/src/services/analytics_service.ts
+++ b/server/src/services/analytics_service.ts
@@ -47,8 +47,11 @@ export class AnalyticsService {
 
   private async _init() {
     this._requester = zmq.socket('pair');
+    let connectionString = 'tcp://127.0.0.1:8002'; // debug mode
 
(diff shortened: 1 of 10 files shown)


server/src/services/analytics_service.ts
```diff
@@ -39,53 +43,79 @@ export class AnalyticsService {
...
+    console.log('Ok');
   }
 
   public get ready(): boolean { return this._ready; }
 
   private async _init() {
     this._requester = zmq.socket('pair');
+    let productionMode = process.env.NODE_ENV !== 'development';
+
+    this._zmqConnectionString = 'tcp://127.0.0.1:8002'; // debug mode
```
rozetko: Maybe we should set it in config? Because there is a chance that port is used

Issue #97, opened 2018-07-25
zmq basic fs


jonyrock:
I want to make a module which saves files async from analytics on node.js side in nedb

jonyrock:
https://www.pythonsheets.com/notes/python-asyncio.html#asynchronous-context-manager

rozetko:
Done

PR #104, opened 2018-07-30
Zmq basic fs #97

jonyrock:
Makes it possible to save/load files my zmq connectivity


server/package.json
```diff
@@ -18,7 +18,9 @@
     "url": "https://github.com/hastic/hastic-server/issues"
   },
   "homepage": "https://github.com/hastic/hastic-server#readme",
-  "dependencies": {},
+  "dependencies": {
+    "@types/es6-promisify": "^6.0.0"
```
rozetko: Guess it should be in dev-deps?

PR #110, opened 2018-08-21
WIP: Antipeak model

VargBurz:
antipeak model without  reverse dataframe


rozetko:
Not actual because of https://github.com/hastic/hastic-server/pull/123

PR #112, opened 2018-08-24
Fix GET /analyticUnits/status

rozetko:
`GET /analyticUnits/status` always returned 'READY' status.
Make it return actual analytic unit status.

### DIFF
--- a/server/src/routes/analytic_units_router.ts
+++ b/server/src/routes/analytic_units_router.ts
@@ -12,7 +12,7 @@ async function getStatus(ctx: Router.IRouterContext) {
       throw new Error('Cannot get status of undefined id');
     }
 
-    let analyticUnit: AnalyticUnit.AnalyticUnit = await AnalyticUnit.findById(analyticUnitId);
+    let analyticUnit = await AnalyticUnit.findById(analyticUnitId);
 
     ctx.response.body = {
       status: analyticUnit.status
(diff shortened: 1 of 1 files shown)


server/src/routes/analytic_units_router.ts
```diff
@@ -1,18 +1,30 @@
...
   try {
-    ctx.response.body = { status: 'READY', errorMessage: undefined };
+    let id = ctx.request.query.id;
+
+    if(id === undefined) {
+      throw new Error('Cannot get status of undefined id');
+    }
+
+    let unit: AnalyticUnit.AnalyticUnit = await AnalyticUnit.findById(id);
+    let errorMessage = unit.error || '';
```
jonyrock: It make sense only if status "failed", we don`t erase errorMessage then set a different status

server/src/routes/analytic_units_router.ts
```diff
@@ -1,18 +1,30 @@
...
-import { createAnalyticUnitFromObject } from '../controllers/analytics_controller'
+import { createAnalyticUnitFromObject } from '../controllers/analytics_controller';
 
 import * as Router from 'koa-router';
 
 
 async function getStatus(ctx: Router.IRouterContext) {
   try {
-    ctx.response.body = { status: 'READY', errorMessage: undefined };
+    let id = ctx.request.query.id;
```
jonyrock: `let analyticUnitId = ctx.request.query.id`

server/src/routes/analytic_units_router.ts
```diff
@@ -1,18 +1,30 @@
...
+
+    if(id === undefined) {
+      throw new Error('Cannot get status of undefined id');
+    }
+
+    let unit: AnalyticUnit.AnalyticUnit = await AnalyticUnit.findById(id);
```
jonyrock: `let analyticUnit = await AnalyticUnit.findById(id);`

PR #118, opened 2018-08-28
Fix prediction

rozetko:


analytics/analytic_unit_manager.py
```diff
@@ -31,6 +31,7 @@ def ensure_worker(analytic_unit_id, analytic_unit_type) -> AnalyticUnitWorker:
     try:
         payload = task['payload']
         payload['data'] = pd.DataFrame(payload['data'], columns = ['timestamp', 'value'])
+        payload['data']['timestamp'] = pd.to_datetime(payload['data']['timestamp'])
```
jonyrock: please don`t touch original `payload`

server/src/controllers/analytics_controller.ts
```diff
@@ -152,29 +151,40 @@ async function processLearningResult(taskResult: any): Promise<{
...
     id,
     AnalyticsTaskType.PREDICT,
-    { pattern, lastPredictionTime: unit.lastPredictionTime }
+    { pattern, lastPredictionTime: unit.lastPredictionTime, data }
   );
+  console.log(task)
   let result = await runTask(task);
-
-  if(result.status === 'FAILED') {
+  console.log(result)
```
jonyrock: lets remove it

analytics/analytic_unit_manager.py
rozetko: You mean we should use copy of `task['payload']` dict?

PR #124, opened 2018-08-30
Make all models work && add reverse peak model

rozetko:
- Subtract min value from dataset before passing to model
- Rename StepModel -> DropModel
- Use cache to save state in all models
- Return `Segment { 'from': <timestamp>, 'to': <timestamp>}` instead of `Segment { 'from': <index>, 'to': <index>}` in all models
- Integrate new peaks model (from https://github.com/hastic/hastic-server/pull/123)
- Integrate new reverse-peaks model (from https://github.com/hastic/hastic-server/pull/123)
- Refactor: make `predict` method in `Model` not abstract and remove it from all children
- Refactor: add abstract `do_predict` method to models


analytics/analytic_unit_manager.py
```diff
@@ -56,4 +55,10 @@ def ensure_worker(analytic_unit_id, analytic_unit_type) -> AnalyticUnitWorker:
             'error': str(e)
         }
 
+def prepare_data(data):
```
jonyrock: please add a comment what is it for or a better name

jonyrock: I think is should be a converation. Please also add typings here.

analytics/analytic_unit_worker.py
```diff
@@ -13,8 +14,8 @@ def __init__(self, analytic_unit_id: str, detector: detectors.Detector):
         self.analytic_unit_id = analytic_unit_id
         self.detector = detector
 
-    async def do_learn(self, segments: list, data: pd.DataFrame, cache: dict) -> dict:
+    async def do_learn(self, segments: list, data: pd.DataFrame, cache: Optional[dict]) -> dict:
```
jonyrock: I like it

analytics/analytic_unit_worker.py
```diff
@@ -13,8 +14,8 @@ def __init__(self, analytic_unit_id: str, detector: detectors.Detector):
...
         return await self.detector.train(data, segments, cache)
 
-    async def do_predict(self, data: pd.DataFrame, cache: dict) -> dict:
+    async def do_predict(self, data: pd.DataFrame, cache: Optional[dict]) -> dict:
```
jonyrock: you also can add a type "AnalyticUnitCache = dict" to make this "dict" more friendly

analytics/detectors/pattern_detector.py
```diff
@@ -30,23 +33,23 @@ def __init__(self, pattern_type):
...
+        new_cache = self.model.fit(dataframe, segments, cache)
         return {
-            'cache': cache
+            'cache': new_cache
         }
 
-    async def predict(self, dataframe: pd.DataFrame, cache: dict):
-        predicted = await self.model.predict(dataframe, cache)
+    async def predict(self, dataframe: pd.DataFrame, cache: Optional[dict]):
+        predicted = self.model.predict(dataframe, cache)
```
jonyrock: in theory we need to split big dataframe to small chunks and do `yield asyncio.sleep(1)`
but please don`t fix it now

jonyrock: we can just make `#TODO: split and sleep`

analytics/models/drop_model.py
```diff
@@ -76,46 +76,34 @@ def fit(self, dataframe: pd.DataFrame, segments: list, cache: dict) -> dict:
                 convolve_list.append(max(convolve))
 
         if len(confidences) > 0:
-            self.state['confidence'] = min(confidences)
+            self.state['confidence'] = float(min(confidences))
```
jonyrock: just interested why do we need `float` here

rozetko: It is `np.int64` by default and can't be serialized by Python

analytics/analytic_unit_manager.py
rozetko: It makes timestamp column `datetime` type
And subtracts min dataset value from dataset

Issue #137, opened 2018-09-06
Datarange sent to models


rozetko:
We got `getQueryRangeForLearningBySegments` method which calculates datarange to send to models
Currently, if you label segments next to each other - model wouldn't even see data in your dashboard timerange
We should change this behaviour somehow

jonyrock:
I made a video abou that. I think we need to make algorithm for this.

jonyrock:
![image](https://user-images.githubusercontent.com/1022757/45266220-43135280-b460-11e8-9559-dfb8b6fe91a8.png)

jonyrock:
![image](https://user-images.githubusercontent.com/1022757/45266237-82da3a00-b460-11e8-8e15-e47bfd600bc2.png)

its me

PR #143, opened 2018-09-11
Datarange sent to models #137

sankerust:
closes https://github.com/hastic/hastic-server/issues/137

Datarange sent to models now uses the range from last segment - diff, to current date.


server/src/controllers/analytics_controller.ts
```diff
@@ -82,7 +82,7 @@ function getQueryRangeForLearningBySegments(segments: Segment.Segment[]) {
   let to = _.maxBy(segments, s => s.to).to;
   let diff = to - from;
   from -= Math.round(diff);
-  to += Math.round(diff);
+  to = Date.now();
```
jonyrock: I don`t like it, but lets take it for now. It will be refactored by making query without `LIMIT`word in query

Issue #145, opened 2018-09-12
Add a note to README.md that the port should be opened in firewall


sankerust:
We need to add a note to README, that the port, that the user wants to use - should be opened in the firewall on the machine.

User feedback.

PR #148, opened 2018-09-12
Add a note to README.md that the port should be opened in firewall #145

sankerust:
closes https://github.com/hastic/hastic-server/issues/145

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ docker run -d \
   hastic/server
 ```
 
-**Make sure that HASTIC_PORT is opened on your system. Default value is ```8000```.**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,6 +113,8 @@ docker run -d \
   hastic/server
 ```
 
+**Make sure that HASTIC_PORT is opened on your system. Default value is ```8000```.**
```
rozetko: Guess would be better
on your system -> in your firewall
` ```8000``` ` -> ` `8000` `
@jonyrock what do you think?

jonyrock: better )

Issue #147, opened 2018-09-12
Reduce amount of required labeled segments for learning


sankerust:
Currently analytics require at least 2 labeled segments to start the learning process.
Only 1 labeled segment should be enough to allow learning.

User feedback.

jonyrock:
@VargBurz can we learn by one labeled segment ?

VargBurz:
ofc, originally, it was one labeled segment for learning

jonyrock:
@VargBurz @rozetko then put 1 instead of 2

rozetko:
Let's use as datarange then `1st segment start - (now - last segment end)` till `now`

PR #153, opened 2018-09-14
Reduce amount of required labeled segments for learning #147

rozetko:
Closes https://github.com/hastic/hastic-server/issues/147
- remove labeled segments amount check
- change query range algorithm


server/src/controllers/analytics_controller.ts
```diff
@@ -74,14 +74,10 @@ async function runTask(task: AnalyticsTask): Promise<TaskResult> {
  * @param segments labeled segments
  */
 function getQueryRangeForLearningBySegments(segments: Segment.Segment[]) {
-  if(segments.length < 2) {
```
jonyrock: lets keep ` if(segments.length === 0) `

Issue #158, opened 2018-09-16
Integrate grafana-datasource-kit


jonyrock:
* add dep (developement)
* refactor necessary code

PR #159, opened 2018-09-16
Integrate grafana datasource kit #158

jonyrock:
There is a bug because interface of `queryByMetric` is changed

This is from the library:
```
/**
 * @param metric to query to Grafana
 * @returns [time, value][] array
 */
export async function queryByMetric(
  metric: GrafanaMetric, url: string, from: number, to: number, apiKey: string
) {
```

I consider this as a bug of `grafana-datasource-kit` rather of hastic.


jonyrock:
https://github.com/CorpGlory/grafana-datasource-kit/issues/11

PR #188, opened 2018-10-04
Decouple processes to different docker containers#187(WIP)

amper43:
- added separate docker files
- added compose file for run together
- added inside_docker env
- added makefile for testin purposes


server/src/services/analytics_service.ts
```diff
@@ -165,7 +171,7 @@ export class AnalyticsService {
 
   private async _onAnalyticsDown() {
     console.log('Analytics is down');
-    if(process.env.NODE_ENV !== 'development') {
+    if(process.env.NODE_ENV !== 'development' && process.env.INSIDE_DOCKER !== 'true') {
```
rozetko: Let's check for env existence instead of its value: `process.env.INSIDE_DOCKER !== 'true'` -> `process.env.INSIDE_DOCKER !== undefined`

analytics/.dockerignore
```diff
@@ -0,0 +1,5 @@
+.git
+npm-debug
+node_modules
+__pychache__
```
rozetko: guess you meant `__pycache__`

server/src/services/analytics_service.ts
amper43: fixed

analytics/.dockerignore
amper43: fixed

PR #204, opened 2018-10-25
Old segments are not deleted #198

### LINKED ISSUE #198: Old segments are not deleted
When you've deleted some segment - prediction starts again
And if some of already found segments are not found second time - they're still kept in db
We should check old found segments and delete them if they're unwanted

amper43:
fixed #198 
image

image

image

### DIFF
--- a/server/src/controllers/analytics_controller.ts
+++ b/server/src/controllers/analytics_controller.ts
@@ -190,10 +190,13 @@ export async function runPredict(id: AnalyticUnit.AnalyticUnitId) {
     //   }
     // }
 
-    let segmentsToRemove = segments.filter(s => function(s) {
+    let predictedSegments = await Segment.findMany(id, { labeled: false });
+    let segmentsToRemove = predictedSegments.filter(s => function(s) {
       return s in payload.segments ? false : true;
     });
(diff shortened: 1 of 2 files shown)


server/src/controllers/analytics_controller.ts
```diff
@@ -190,6 +190,12 @@ export async function runPredict(id: AnalyticUnit.AnalyticUnitId) {
     //   }
     // }
 
+    let predictedSegments = await Segment.findMany(id, { labeled: false });
+    let segmentsToRemove = predictedSegments.filter(s => function(s) {
```
rozetko: It would better be `let segmentsToRemove = predictedSegments.filter(s => payload.segments.indexOf(s) > -1 ? false : true)`

server/src/controllers/analytics_controller.ts
```diff
@@ -190,6 +190,12 @@ export async function runPredict(id: AnalyticUnit.AnalyticUnitId) {
     //   }
     // }
 
+    let predictedSegments = await Segment.findMany(id, { labeled: false });
```
rozetko: You're deleting deleted segments along with predicted

PR #217, opened 2018-10-28
Analytics works with one labeled segment#191/212

VargBurz:
problem was: function 'get_av_model' applied to nonexistent element
issue 191
issue 212

### DIFF
--- a/analytics/utils/__init__.py
+++ b/analytics/utils/__init__.py
@@ -212,8 +212,11 @@ def ar_mean(numbers):
 
 def get_av_model(patterns_list):
     x = len(patterns_list[0])
-    if len(patterns_list[1]) != x:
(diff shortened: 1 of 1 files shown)


analytics/utils/__init__.py
```diff
@@ -212,10 +212,9 @@ def ar_mean(numbers):
 
 def get_av_model(patterns_list):
     x = len(patterns_list[0])
-    try:
-        if len(patterns_list[1]) != x:
-            raise NameError('All elements of patterns_list should have same length')
-    except IndexError:
+    if len(pattern_list) > 1 and len(patterns_list[1]) != x:
```
jonyrock: I believe that `len(pattern_list) > 1` is always `true` if I can go thought the line ` x = len(patterns_list[0])` which is above.

PR #218, opened 2018-10-28
Models with filtering#186

### LINKED ISSUE #186: Incorrect filtration of detected segments
in first filtration delete all close segments except the first

VargBurz:
closes #186 

add new logic of first filtration in all models, except general:
if there are few close pattern, analytics choose the highest or lowest one.


analytics/models/peak_model.py
```diff
@@ -114,16 +114,10 @@ def do_predict(self, dataframe: pd.DataFrame):
...
-        for i in range(1, len(segments)):
-            if segments[i] < segments[i - 1] + variance_error:
-                delete_list.append(segments[i])
-        for item in delete_list:
-            segments.remove(item)
-
-        delete_list = []
+        variance_error = self.state['WINDOW_SIZE']
+        close_patterns = utils.close_filtering(segments, variance_error)
+        segments = utils.best_pat(close_patterns, data, "max")
```
rozetko: Maybe we can create function in utils which takes `segments, variance_error, data, 'max'` and returns best patterns? Which uses `close_filtering` inside

analytics/models/trough_model.py
```diff
@@ -115,16 +115,9 @@ def do_predict(self, dataframe: pd.DataFrame):
...
-        for i in range(1, len(segments)):
-            if segments[i] < segments[i - 1] + variance_error:
-                delete_list.append(segments[i])
-        for item in delete_list:
-            segments.remove(item)
-
-        delete_list = []
+        variance_error = self.state['WINDOW_SIZE']
+        close_patterns = utils.close_filtering(segments, variance_error)
+        segments = utils.best_pat(close_patterns, data, "min")
```
rozetko: btw, we don't use double quotes: `"` -> `'`

analytics/models/drop_model.py
```diff
@@ -154,17 +154,9 @@ def do_predict(self, dataframe: pd.DataFrame) -> list:
...
-        for i in range(1, len(segments)):
-            if segments[i] < segments[i - 1] + variance_error:
-                delete_list.append(segments[i])
-        # for item in delete_list:
-        #    segments.remove(item)
-        delete_list = []
-
+        variance_error = self.state['WINDOW_SIZE']
+        close_patterns = utils.close_filtering(segments, variance_error)
+        segments = utils.best_pat(close_patterns, data, "min")
```
rozetko: The same thing here

analytics/utils/__init__.py
```diff
@@ -221,3 +221,35 @@ def get_av_model(patterns_list):
             av_val.append(j.values[i])
         model_pat.append(ar_mean(av_val))
     return model_pat
+
+def close_filtering(pat_list, win_size):
+    s = []
+    s.append([])
+    s[0].append(pat_list[0])
```
rozetko: Why don't change first three lines to `s = [[pat_list[0]]]`?

analytics/utils/__init__.py
```diff
@@ -221,3 +221,35 @@ def get_av_model(patterns_list):
...
+    k = 0
+    for i in range(1, len(pat_list)):
+        if pat_list[i] - win_size <= s[k][-1]:
+            s[k].append(pat_list[i])
+        else:
+            k += 1
+            s.append([])
+            s[k].append(pat_list[i])
```
rozetko: why not `s.append([pat_list[i]])`?

analytics/utils/__init__.py
```diff
@@ -221,3 +221,35 @@ def get_av_model(patterns_list):
...
+    return s
+
+def best_pat(pat_list, data, dir):
```
rozetko: Can be simpler:
```python
def best_pat(pat_list, data, dir):
  func = np.argmax if dir == 'max' else np.argmin
  return [func(val) for val in pat_list]
```
Didn't test it though, but seems working
@amper43 can you check it please?

analytics/models/peak_model.py
VargBurz: I suggest to keep two functions, because of diffrent logic. Besides, `best_pat` will  be changed and/or will be add new metod for few models.

analytics/models/trough_model.py
VargBurz: done

analytics/models/drop_model.py
VargBurz: done

analytics/utils/__init__.py
VargBurz: image

analytics/utils/__init__.py
VargBurz: done

analytics/utils/__init__.py
VargBurz: done

analytics/utils/__init__.py
rozetko: You're right
Get function wrong

PR #226, opened 2018-11-01
models_with_filtering_v2#186

VargBurz:
new PR with filtering


amper43:
dupliction of https://github.com/hastic/hastic-server/pull/218

Issue #215, opened 2018-10-27
Analytics project structure improvement


jonyrock:
I think we should rethink our mess with docs/files in analytics. 

I nice start is: https://stackoverflow.com/questions/193161/what-is-the-best-project-structure-for-a-python-application

jonyrock:
@rozetko I think I also need to check 
* node.js part
* the process of building python bundle of analytics

what else?

rozetko:
- paths in `config.py`
- imports

rozetko:
Done

PR #228, opened 2018-11-02
Analytics project structure improvement #215

jonyrock:
* bin folder with server
* analytics folder
* import paths fix
* node.js path to analytics fix


jonyrock:
@amper43 please take a note https://github.com/hastic/hastic-server/issues/229

analytics/Dockerfile
```diff
@@ -17,4 +17,4 @@ VOLUME [ "/var/www/data" ]
 
 COPY . /var/www/analytics/
 
-CMD ["python", "server.py"]
+CMD ["python", "bin/server"]
```
amper43: maybe it will be better to  make `server.py` executable and run it directly?

Issue #207, opened 2018-10-25
Error: min() arg in an empty sequence


amper43:
![2](https://user-images.githubusercontent.com/5675912/47513298-9c0b3e80-d886-11e8-9af8-943c4c441243.jpg)

amper43:
2018-10-26 16:36:54,265 [Analytics] [ERROR]  handle_analytic_task exception: 'Traceback (most recent call last):
  File "/home/jo/extra/hastic-server/analytics/analytic_unit_manager.py",line 35, in handle_analytic_task
    result_payload = await worker.do_learn(payload['segments'], data, payload['cache'])
  File "/home/jo/extra/hastic-server/analytics/analytic_unit_worker.py", line 19, in do_learn
    return await self.detector.train(data, segments, cache)
  File "/home/jo/extra/hastic-server/analytics/detectors/pattern_detector.py", line 40, in train
    new_cache = self.model.fit(dataframe, segments, cache)
  File "/home/jo/extra/hastic-server/analytics/models/model.py", line 34,in fit
    self.do_fit(dataframe, segments)
  File "/home/jo/extra/hastic-server/analytics/models/jump_model.py", line 50, in do_fit
    min_jump = min(flat_segment_dropna)
ValueError: min() arg is an empty sequence

sankerust:
I was able to reproduce same bug on same `Jumps` model.

jonyrock:
@VargBurz please add comments why it happened in free form. It is necessary for closing task.

amper43:
will be fixed by https://github.com/hastic/hastic-server/issues/231

PR #230, opened 2018-11-02
Tests for Error min arg in an empty sequence #207

amper43:
tests for #231


analytics/utils/common.py
```diff
@@ -212,7 +212,11 @@ def ar_mean(numbers):
 
 def get_av_model(patterns_list):
     x = len(patterns_list[0])
+<<<<<<< HEAD
```
rozetko: The same thing again :(

analytics/tests/test_dataset.py
```diff
@@ -0,0 +1,24 @@
...
+from models.jump_model import JumpModel
+
+class TestDataset(unittest.TestCase):
+
+    def test_models_with_corrupted_dataframe(self):
+        data = [[1523889000000 + i, float('nan')] for i in range(10)]
+        dataframe = pd.DataFrame(data, columns=['timestamp', 'value'])
+        segments = []
+
+        model = JumpModel()
```
rozetko: It's not a jump model test -> maybe move code you test from specific model to utils?

PR #235, opened 2018-11-06
Incorrect work of models with single patterns #234

### LINKED ISSUE #234: Incorrect work of models with single patterns
The model removes all possible patterns, if the distance between them is bigger than `WINDOW_SIZE`.
Steps to reproduce: 
- Use any dataset with single patterns, for example peaks.
- Peaks in dataset should be narrow and far from each other.
Solution:
Change filtering and utils.

VargBurz:
Fixes #234 
add first value of patterns list = val[0] instead 0


VargBurz:
fixes #234

PR #236, opened 2018-11-11
Proccesing for nan values #231

### LINKED ISSUE #231: Processing for 'NaN' values
When there is no data for requested segment, returned values fills by 'NaN'. Analytics must processing this case correctly. Offered solution: check that non-NaN data fills segment a least on 10%.

VargBurz:
close #231


analytics/analytics/utils/common.py
```diff
@@ -250,3 +250,19 @@ def best_pat(pat_list, data, dir):
...
     return new_pat_list
+
+def nan_checker(segment):
+    nan_list = np.isnan(segment)
+    nan_count = 0
+    nan_indexes = []
+    for i, val in enumerate(nan_list):
+        if val:
+            nan_count += 1
+            nan_inexes.append(i)
```
```suggestion
            nan_indexes.append(i)
```

analytics/analytics/utils/common.py
```diff
@@ -250,3 +250,19 @@ def best_pat(pat_list, data, dir):
...
+    return nan_count / len(segment), nan_indexes
+
+def nan_for_zero(segment, nan_list):
```
rozetko: You can use `numpy.nan_to_num` method which does the same thing
https://docs.scipy.org/doc/numpy/reference/generated/numpy.nan_to_num.html

analytics/analytics/models/general_model.py
```diff
@@ -36,9 +36,9 @@ def do_fit(self, dataframe: pd.DataFrame, segments: list) -> None:
             if segment['labeled']:
                 segment_from_index = utils.timestamp_to_index(dataframe, pd.to_datetime(segment['from'], unit='ms'))
                 segment_to_index = utils.timestamp_to_index(dataframe, pd.to_datetime(segment['to'], unit='ms'))
-
                 segment_data = data[segment_from_index: segment_to_index + 1]
-                if len(segment_data) == 0:
+                percent_of_nans = utils.nan_checker(segment_data)[0]
```
rozetko: Can be just `percent_of_nans = segment_data.count(np.NaN) / len(segment_data)`

analytics/analytics/utils/common.py
```diff
@@ -250,3 +250,19 @@ def best_pat(pat_list, data, dir):
...
+    nan_list = np.isnan(segment)
+    nan_count = 0
+    nan_indexes = []
+    for i, val in enumerate(nan_list):
+        if val:
+            nan_count += 1
+            nan_indexes.append(i)
+    return nan_count / len(segment), nan_indexes
+
+def nan_for_zero(segment, nan_list):
```
```suggestion
def nan_to_zero(segment, nan_list):
```

analytics/analytics/models/peak_model.py
```diff
@@ -125,6 +126,14 @@ def __filter_prediction(self, segments: list, data: list) -> list:
             if segment > self.state['WINDOW_SIZE']:
                 convol_data = data[segment - self.state['WINDOW_SIZE']: segment + self.state['WINDOW_SIZE'] + 1]
                 convol_data = convol_data - min(convol_data)
+                percent_of_nans = utils.nan_checker(convol_data)[0]
+                if percent_of_nans > 0.5:
+                    delete_list.append(segment)
+                    continue
+                elif 0 < percent_of_nans < 0.5:
```
rozetko: What if `percent_of_nans == 0.5` or `percent_of_nans == 0`?

rozetko: Maybe just `else` would be better?

VargBurz: If `percent_of_nans == 0`  it will be common case - nothing should change.

PR #281, opened 2018-12-03
Add some sort of verification of incoming and outgoing data in analytics #227

### LINKED ISSUE #227: Add some sort of verification of incoming and outgoing data in analytics
need to add some varification for catching the trash information, like NaN. or empty data in input.
And NaN in output

VargBurz:
closes #227

### DIFF
--- a/analytics/analytics/utils/common.py
+++ b/analytics/analytics/utils/common.py
@@ -275,7 +275,13 @@ def find_nan_indexes(segment: pd.Series) -> list:
             nan_indexes.append(i)
     return nan_indexes
 
-def nan_to_zero(segment: pd.Series, nan_list: list) -> pd.Series:
+def check_nan_values(segment: [pd.Series, list]) -> [pd.Series, list]:
+    nan_list = utils.find_nan_indexes(segment)
+    if len(nan_list) > 0:
+        segment = utils.nan_to_zero(segment, nan_list)
+    return segment
+
+def nan_to_zero(segment: [pd.Series, list], nan_list: list) -> [pd.Series, list]:
     if type(segment) == pd.Series:
         for val in nan_list:
             segment.values[val] = 0
@@ -285,9 +291,7 @@ def nan_to_zero(segment: pd.Series, nan_list: list) -> pd.Series:
     return segment
 
 def find_confidence(segment: pd.Series) -> float:
-    nan_list = utils.find_nan_indexes(segment)
-    if len(nan_list) > 0:
-        segment = utils.nan_to_zero(segment, nan_list)
+    segment = utils.check_nan_values(segment)
     segment_min = min(segment)
     segment_max = max(segment)
     return 0.2 * (segment_max - segment_min)
@@ -297,8 +301,8 @@ def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
     right_bound = center + window_size + 1
     if left_bound < 0:
         left_bound = 0
-    if right_bound > len(data) + 1:
-        right_bound = len(data) + 1
+    if right_bound > len(data):
+        right_bound = len(data)
     return data[left_bound: right_bound]
 
 def subtract_min_without_nan(segment: list) -> list:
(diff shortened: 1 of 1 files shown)


analytics/analytics/utils/common.py
```diff
@@ -267,32 +267,48 @@ def best_pat(pat_list, data, dir):
...
     segment_min = min(segment)
     segment_max = max(segment)
     return 0.2 * (segment_max - segment_min)
 
 def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
     left_bound = center - window_size
     right_bound = center + window_size + 1
+    if left_bound < 0:
+        left_bound = 0
+    if right_bound > len(data) + 1:
```
jonyrock:
```suggestion
    if right_bound > len(data):
```

analytics/analytics/utils/common.py
```diff
@@ -267,32 +267,48 @@ def best_pat(pat_list, data, dir):
...
+        right_bound = len(data) + 1
```
jonyrock:
```suggestion
        right_bound = len(data)
```

analytics/analytics/utils/common.py
```diff
@@ -267,32 +267,48 @@ def best_pat(pat_list, data, dir):
...
 def subtract_min_without_nan(segment: list) -> list:
-    if not np.isnan(min(segment)):
-        segment = segment - min(segment)
+    if len(segment) == 0:
+        return []
+    nan_list = utils.find_nan_indexes(segment)
+    if len(nan_list) > 0:
+        return segment
+    else:
+        segment = segment - min(segment)
```
jonyrock: what type of `min(segment) ` ?

analytics/analytics/utils/common.py
```diff
@@ -267,32 +267,48 @@ def best_pat(pat_list, data, dir):
...
             nan_indexes.append(i)
     return nan_indexes
 
-def nan_to_zero(segment, nan_list):
-    for val in nan_list:
-        segment[val] = 0
+def nan_to_zero(segment: pd.Series, nan_list: list) -> pd.Series:
+    if type(segment) == pd.Series:
+        for val in nan_list:
+            segment.values[val] = 0
```
rozetko: What if there is no element with `val` index in segment?

analytics/analytics/utils/common.py
```diff
@@ -267,32 +267,48 @@ def best_pat(pat_list, data, dir):
...
+    else:
+        for val in nan_list:
+            segment[val] = 0
     return segment
 
 def find_confidence(segment: pd.Series) -> float:
+    nan_list = utils.find_nan_indexes(segment)
+    if len(nan_list) > 0:
+        segment = utils.nan_to_zero(segment, nan_list)
```
amper43: Maybe it would be better to use https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.fillna.html

rozetko: There are the same 3 lines in `get_convolve`
Maybe we should create function?

analytics/analytics/utils/common.py
```diff
@@ -267,32 +267,48 @@ def best_pat(pat_list, data, dir):
...
     for i, val in enumerate(nan_list):
         if val:
             nan_indexes.append(i)
     return nan_indexes
 
-def nan_to_zero(segment, nan_list):
-    for val in nan_list:
-        segment[val] = 0
+def nan_to_zero(segment: pd.Series, nan_list: list) -> pd.Series:
+    if type(segment) == pd.Series:
```
rozetko: If you look at arguments' types - `segment` can be only `pd.Series`
Let's specify other possible types explicitly
e.g. `Union[pd.Series, list]`

analytics/analytics/utils/common.py
```diff
@@ -267,32 +267,48 @@ def best_pat(pat_list, data, dir):
...
     return 0.2 * (segment_max - segment_min)
 
 def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
     left_bound = center - window_size
     right_bound = center + window_size + 1
+    if left_bound < 0:
+        left_bound = 0
+    if right_bound > len(data) + 1:
+        right_bound = len(data) + 1
     return data[left_bound: right_bound]
```
```suggestion
    return data[left_bound : right_bound]
```

analytics/analytics/utils/common.py
VargBurz: mostly `float`

analytics/analytics/utils/common.py
VargBurz: Sometimes this method used for `list` data, not only `pd.Series`.

PR #295, opened 2018-12-06
Update Data Puller state #294

### LINKED ISSUE #294: Update Data Puller state
Currently DataPuller has its own state which is not connected to db
It's better to store only ids of analytic units and pull them from db each tick

### LINKED ISSUE #296: Store only actual analytic units in Data Puller
Currently units are added to DataPuller only on unit creation
Thus, when you are restarting server - DataPuller has no units
Maybe we can just get units with `alert === true` from db and push / pull data for them?
@jonyrock @amper43 what do you think?

rozetko:
Fixes #294 
Fixes #296

- Get analytic units from DB on each puller tick

### DIFF
--- a/server/src/services/data_puller.ts
+++ b/server/src/services/data_puller.ts
@@ -8,7 +8,7 @@ import { queryByMetric } from 'grafana-datasource-kit';
 import * as _ from 'lodash';
 
 
-type Data = { values: [number, number][], columns: string[] };
+type MetricDataChunk = { values: [number, number][], columns: string[] };
 
 const PULL_PERIOD_MS = 5000;
 
--- a/server/src/services/data_puller.ts
+++ b/server/src/services/data_puller.ts
@@ -28,7 +28,9 @@ export class DataPuller {
     delete this._unitTimes[id];
   }
 
-  private pullData(unit: AnalyticUnit.AnalyticUnit, from: number, to: number) {
+  private async pullData(unit: AnalyticUnit.AnalyticUnit, from: number, to: number): Promise<{
+    values: [number, number][]; columns: string[];
+  }> {
     if(unit === undefined) {
       throw Error(`puller: can't pull undefined unit`);
     }
--- a/server/src/services/data_puller.ts
+++ b/server/src/services/data_puller.ts
@@ -83,7 +83,7 @@ export class DataPuller {
   }
 
   async * getDataGenerator(analyticUnit: AnalyticUnit.AnalyticUnit, duration: number):
-    AsyncIterableIterator<Data> {
+    AsyncIterableIterator<MetricDataChunk> {
 
     const getData = async () => {
       try {
--- a/server/src/services/data_puller.ts
+++ b/server/src/services/data_puller.ts
@@ -28,7 +28,7 @@ export class DataPuller {
     }
   }
 
-  private async pullData(unit: AnalyticUnit.AnalyticUnit, from: number, to: number): Promise<Data> {
+  private async pullData(unit: AnalyticUnit.AnalyticUnit, from: number, to: number): Promise<MetricDataChunk> {
     if(unit === undefined) {
       throw Error(`puller: can't pull undefined unit`);
     }
--- a/server/src/services/data_puller.ts
+++ b/server/src/services/data_puller.ts
@@ -7,25 +7,18 @@ import { queryByMetric } from 'grafana-datasource-kit';
 
 import * as _ from 'lodash';
 
-
-declare type UnitTime = {
-  unit: AnalyticUnit.AnalyticUnit,
-  time: number
-};
-
 export class DataPuller {
 
-  private PULL_PERIOD_MS: number = 5000;
-  private _interval: number = 1000;
+  private PULL_PERIOD_MS = 5000;
+  private _interval = 1000;
   private _timer: any = null;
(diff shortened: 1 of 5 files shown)


server/src/services/data_puller.ts
```diff
@@ -7,33 +7,26 @@ import { queryByMetric } from 'grafana-datasource-kit';
...
-declare type UnitTime = {
-  unit: AnalyticUnit.AnalyticUnit,
-  time: number
-};
-
 export class DataPuller {
 
-  private PULL_PERIOD_MS: number = 5000;
-  private _interval: number = 1000;
+  private PULL_PERIOD_MS = 5000;
```
jonyrock: why not put `:number` here?

jonyrock: please choose one of this options: https://stackoverflow.com/questions/37265275/how-to-implement-class-constants-in-typescript

I personally prefer to put constants outside of the class

server/src/services/data_puller.ts
```diff
@@ -7,33 +7,26 @@ import { queryByMetric } from 'grafana-datasource-kit';
...
-    let unitTime: UnitTime = {unit, time };
-    this._unitTimes[unit.id] = unitTime;
+    this._unitTimes[unit.id] = time;
   }
 
   public deleteUnit(id: AnalyticUnit.AnalyticUnitId) {
     delete this._unitTimes[id];
   }
 
   private pullData(unit: AnalyticUnit.AnalyticUnit, from: number, to: number) {
```
jonyrock: can we add return type here, please

server/src/services/data_puller.ts
```diff
@@ -73,18 +65,19 @@ export class DataPuller {
 
     let now = Date.now();
 
-    _.forOwn(this._unitTimes, async value => {
-      if(!value.unit.alert) {
+    _.forOwn(this._unitTimes, async (time: number, analyticUnitId: AnalyticUnit.AnalyticUnitId) => {
```
jonyrock: you execute `_.forOwn` callback one after another. You should pull data for each analyticUnit separately. So you need to run make workers for each puller.

server/src/services/data_puller.ts
jonyrock: Please see https://github.com/hastic/hastic-grafana-graph-panel/blob/master/src/services/analytic_service.ts#L89

rozetko: There is no sence because typescript will inference it by itself

server/src/services/data_puller.ts
```diff
@@ -46,46 +45,65 @@ export class DataPuller {
...
-    this._timer = setTimeout(this.puller.bind(this), this._interval);
+      const now = Date.now();
+      let payload = { data, from: time, to: now };
+      this._unitTimes[analyticUnit.id] = now;
+      this.pushData(analyticUnit, payload);
+    }
+  }
+
+  async * getDataGenerator(analyticUnit: AnalyticUnit.AnalyticUnit, duration: number):
+    AsyncIterableIterator<{ values: [number, number][]; columns: string[]; }> {
```
jonyrock: Please make a new type `{ values: [number, number][]; columns: string[]; }` for this expressoin

server/src/services/data_puller.ts
```diff
@@ -8,32 +8,28 @@ import { queryByMetric } from 'grafana-datasource-kit';
 import * as _ from 'lodash';
 
 
-declare type UnitTime = {
-  unit: AnalyticUnit.AnalyticUnit,
-  time: number
-};
+type Data = { values: [number, number][], columns: string[] };
```
jonyrock: Can we be a little bit more specific than just `Data`, please?

server/src/services/data_puller.ts
```diff
@@ -8,32 +8,28 @@ import { queryByMetric } from 'grafana-datasource-kit';
...
-    delete this._unitTimes[id];
+  public deleteUnit(analyticUnitId: AnalyticUnit.AnalyticUnitId) {
+    if(_.has(this._unitTimes, analyticUnitId)) {
+      delete this._unitTimes[analyticUnitId];
+    }
   }
 
-  private pullData(unit: AnalyticUnit.AnalyticUnit, from: number, to: number) {
-    if(!unit) {
+  private async pullData(unit: AnalyticUnit.AnalyticUnit, from: number, to: number): Promise<Data> {
```
jonyrock:
```suggestion
  private async pullData(unit: AnalyticUnit.AnalyticUnit, from: number, to: number): Promise<MetricDataChunk> {
```

server/src/services/data_puller.ts
```diff
@@ -48,46 +44,65 @@ export class DataPuller {
...
-    this._timer = setTimeout(this.puller.bind(this), this._interval);
+      const now = Date.now();
+      let payload = { data, from: time, to: now, pattern: analyticUnit.type };
+      this._unitTimes[analyticUnit.id] = now;
+      this.pushData(analyticUnit, payload);
+    }
+  }
+
+  async * getDataGenerator(analyticUnit: AnalyticUnit.AnalyticUnit, duration: number):
+    AsyncIterableIterator<Data> {
```
rozetko:
```suggestion
    AsyncIterableIterator<MetricDataChunk> {
```

PR #305, opened 2018-12-07
DataPuller: wrong from and to #304

### LINKED ISSUE #304: DataPuller: wrong from and to
If there is `lastDetectionTime` for `AnalyticUnit` and it is in nanoseconds - DataPuller still trying to query datasource-kit

It`s because we receive `lastDetectionTime` from analytics in nanoseconds

rozetko:
Fixes #304 

Convert lastDetectionTime from ns to ms

### DIFF
--- a/analytics/analytics/detectors/pattern_detector.py
+++ b/analytics/analytics/detectors/pattern_detector.py
@@ -53,7 +53,8 @@ class PatternDetector(Detector):
         newCache = detected['cache']
 
         last_dataframe_time = dataframe.iloc[-1]['timestamp']
-        last_detection_time = last_dataframe_time.value
+        # TODO: convert from ns to ms more proper way (not dividing by 10^6)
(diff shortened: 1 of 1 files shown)


analytics/analytics/detectors/pattern_detector.py
```diff
@@ -53,7 +53,8 @@ def detect(self, dataframe: pd.DataFrame, cache: Optional[models.ModelCache]) ->
         newCache = detected['cache']
 
         last_dataframe_time = dataframe.iloc[-1]['timestamp']
-        last_detection_time = last_dataframe_time.value
+        # TODO: convert from ns to ms more proper way (not dividing by 10^6)
```
jonyrock:
```suggestion
        # TODO: convert from nanoseconds to millisecond in a better way (not dividing by 10^6)
```

PR #319, opened 2018-12-12
Analytics testing #210

### LINKED ISSUE #210: Analytics testing
- [x] Choose a testing framework
- [x] Generate test datasets for each model
- [ ] Add basic tests for each model
  - [ ] For predicting
  - [ ] For anti-segments

VargBurz:
closes #210

### DIFF
--- a/analytics/analytics/models/model.py
+++ b/analytics/analytics/models/model.py
@@ -31,7 +31,10 @@ class Model(ABC):
 
                 segment_length = abs(segment_to_index - segment_from_index)
                 segment_length_list.append(segment_length)
-        self.state['WINDOW_SIZE'] = math.ceil(max(segment_length_list) / 2)
+        if len(segment_length_list) > 0:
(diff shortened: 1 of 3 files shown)


analytics/analytics/models/model.py
```diff
@@ -31,7 +31,10 @@ def fit(self, dataframe: pd.DataFrame, segments: list, cache: Optional[ModelCach
 
                 segment_length = abs(segment_to_index - segment_from_index)
                 segment_length_list.append(segment_length)
-        self.state['WINDOW_SIZE'] = math.ceil(max(segment_length_list) / 2)
+        if len(segment_length_list) > 0:
```
amper43: this if - else can be more simple

`self.state['WINDOW_SIZE'] = math.ceil(max(segment_length_list) / 2) if len(segment_length_list) else 0`

VargBurz: I think that all if -else must be in the same format.

jonyrock: I am fine with original @VargBurz 's version

PR #322, opened 2018-12-14
Error name parse segment is not defined #315

### LINKED ISSUE #315: Error: name 'parse_segment' is not defined
image

Steps to reproduce:
* create "Peaks" analyticUnit
* label a segment
* save
* delete segment 
* save
* label a new overlapping with deleted segment, but not entirely
* save

Should have result as on screenshot above.

Server info:
image

VargBurz:
closes #315

### DIFF
--- a/analytics/tests/test_dataset.py
+++ b/analytics/tests/test_dataset.py
@@ -78,7 +78,7 @@ class TestDataset(unittest.TestCase):
         data = {'timestamp': data_ind, 'value': data_val}
         dataframe = pd.DataFrame(data)
         dataframe['timestamp'] = pd.to_datetime(dataframe['timestamp'], unit='ms')
-        segments = [{'_id': 'Esl7uetLhx4lCqHa', 'analyticUnitId': 'opnICRJwOmwBELK8', 'from': 1523889000010, 'to': 1523889000012, 'labeled': True, 'deleted': False},
+        segments = [{'_id': 'Esl7uetLhx4lCqHa', 'analyticUnitId': 'opnICRJwOmwBELK8', 'from': 1523889000009, 'to': 1523889000011, 'labeled': True, 'deleted': False},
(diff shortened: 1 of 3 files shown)


analytics/tests/test_dataset.py
```diff
@@ -23,6 +23,86 @@ def test_models_with_corrupted_dataframe(self):
...
+                    {'_id': 'Esl7uetLhx4lCqHa', 'analyticUnitId': 'opnICRJwOmwBELK8', 'from': 1523889000002, 'to': 1523889000008, 'labeled': False, 'deleted': True}]
+
+        try:
+            model = models.DropModel()
+            model_name = model.__class__.__name__
+            model.fit(dataframe, segments, dict())
+        except ValueError:
+            self.fail('Model {} raised unexpectedly'.format(model_name))
+
+    def test_general_antisegments(self):
```
rozetko: There are a lot of duplicating code
The only difference is model type, as far as I see
You can do the same thing as here (create instances of each model and then test them in a loop):
https://github.com/hastic/hastic-server/blob/master/analytics/tests/test_dataset.py#L13-L25

VargBurz: I want to have tests for each model. And there are diffrent data values..

analytics/tests/test_dataset.py
```diff
@@ -23,6 +23,86 @@ def test_models_with_corrupted_dataframe(self):
...
+        data_val = [1.0, 2.0, 1.0, 2.0, 5.0, 6.0, 3.0, 2.0, 1.0, 1.0, 8.0, 9.0, 8.0, 1.0, 2.0, 3.0, 2.0, 1.0, 1.0, 2.0]
+        data_ind = [1523889000000 + i for i in range(20)]
```
jonyrock:
```suggestion
        data_ind = [1523889000000 + i for i in range(len(data_val))]
```

analytics/tests/test_dataset.py
rozetko: Why do you need different datasets? You're just testing to say whether `fit` fails or not

VargBurz: because in diffrent models need to send diffrent data.
and it is uncomfortable to use one test for all models (for me)

PR #329, opened 2018-12-19
Endpoint for analytic unit types #328

### LINKED ISSUE #328: Endpoint for analytic unit types
We should create an endpoint like `GET /analyticUnits/types` with response like:
```
{
  patterns: [{ name: 'peak', ... }, ...],
  threshold: [{ name: 'threshold' }]
}
```
Where `patterns` and `threshold` are detectors' names

It is needed to implement https://github.com/hastic/hastic-grafana-app/issues/112

rozetko:
Closes #328 
`GET /analyticUnits/types`: return const types list

### DIFF
--- a/server/src/routes/analytic_units_router.ts
+++ b/server/src/routes/analytic_units_router.ts
@@ -89,16 +89,7 @@ async function getUnits(ctx: Router.IRouterContext) {
 }
 
 function getTypes(ctx: Router.IRouterContext) {
-  try {
-    ctx.response.body = AnalyticUnit.ANALYTIC_UNIT_TYPES;
-  } catch(e) {
-    console.error(e);
-    ctx.response.status = 404;
-    ctx.response.body = {
-      code: 404,
-      message: `GET /analyticUnits/types error: ${e.message}`
-    };
-  }
+  ctx.response.body = AnalyticUnit.ANALYTIC_UNIT_TYPES;
 }
 
 async function createUnit(ctx: Router.IRouterContext) {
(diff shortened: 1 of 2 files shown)


server/src/routes/analytic_units_router.ts
```diff
@@ -88,6 +88,52 @@ async function getUnits(ctx: Router.IRouterContext) {
...
+        },
+        {
+          name: 'Jumps',
+          value: 'JUMP'
+        },
+        {
+          name: 'Drops',
+          value: 'DROP'
+        },
+        {
```
jonyrock: lets remove `custom` for now cuz we don`t really support this feature yet

server/src/routes/analytic_units_router.ts
```diff
@@ -88,6 +88,52 @@ async function getUnits(ctx: Router.IRouterContext) {
   }
 }
 
+function getTypes(ctx: Router.IRouterContext) {
+  try {
+    ctx.response.body = {
+      pattern: [
```
jonyrock: Can we move this object to some kind of model or config or constant, not here.

server/src/routes/analytic_units_router.ts
```diff
@@ -88,6 +88,52 @@ async function getUnits(ctx: Router.IRouterContext) {
...
+      ],
+      threshold: [
+        {
+          name: 'Threshold',
+          value: 'THRESHOLD'
+        }
+      ]
+    };
+  } catch(e) {
+    console.error(e);
```
jonyrock: What kind of error do you expect here? Looks suspicious to me...

PR #336, opened 2019-01-14
Move data cropping to the models #335

### LINKED ISSUE #335: Move data cropping to the models
It is necessary to move cropping of data that goes to analytics directly inside of the models.
Also we need to add a check for negative values.

VargBurz:
closes #335

### DIFF
--- a/analytics/analytics/models/general_model.py
+++ b/analytics/analytics/models/general_model.py
@@ -28,6 +28,7 @@ class GeneralModel(Model):
         self.all_conv = []
 
     def do_fit(self, dataframe: pd.DataFrame, segments: list) -> None:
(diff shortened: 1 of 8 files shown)


analytics/analytics/models/general_model.py
```diff
@@ -28,6 +28,7 @@ def __init__(self):
         self.all_conv = []
 
     def do_fit(self, dataframe: pd.DataFrame, segments: list) -> None:
+        data = utils.cut_dataframe(dataframe)
         data = dataframe['value']
```
rozetko: You're cutting `dataframe` and assigning result to `data`
But using `dataframe` again after
Guess it should be `data = data['value']`?

PR #355, opened 2019-01-21
Sort segments in process detection result #354

### LINKED ISSUE #354: Sort segments in processDetectionResult
In `analytic_controller.ts` in `processDetectionResult` method we don't sort segments, but use `_.last()` that choose wrong segment for webhook

amper43:
* sort segments in `processDetectionResult`
* send only non-sended last segment

fixes #354

### DIFF
--- a/server/src/controllers/analytics_controller.ts
+++ b/server/src/controllers/analytics_controller.ts
@@ -299,6 +299,7 @@ async function processDetectionResult(analyticUnitId: AnalyticUnit.AnalyticUnitI
       `Missing lastDetectionTime in result or it is corrupted: ${JSON.stringify(detectionResult)}`
     );
   }
+  console.debug(`got detection result with ${detectionResult.segments.length} segments`);
 
   const segments = detectionResult.segments.map(
     segment => new Segment.Segment(analyticUnitId, segment.from, segment.to, false, false)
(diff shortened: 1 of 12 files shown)


server/src/controllers/analytics_controller.ts
```diff
@@ -305,9 +305,19 @@ async function processDetectionResult(analyticUnitId: AnalyticUnit.AnalyticUnitI
       `Missing lastDetectionTime in result or it is corrupted: ${JSON.stringify(detectionResult)}`
     );
   }
-  console.debug(`got detection result with ${detectionResult.segments.length} segments`);
+  console.debug(`got detection result for ${analyticUnitId} with ${detectionResult.segments.length} segments`);
 
-  const segments = detectionResult.segments.map(
+  let sortedSegments: {from, to}[] = detectionResult.segments;
+  sortedSegments = sortedSegments.sort((a, b) => {
```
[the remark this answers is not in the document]
amper43: yes, it would be bettr, thanks

server/src/services/alert_service.ts
```diff
@@ -13,7 +13,17 @@ export class Alert {
   public update(now: number) {};
 }
 
-class PatternAlert extends Alert {};
+class PatternAlert extends Alert {
+
+  private lastSendedSegment: Segment;
```
rozetko: `lastSentSegment`?:)

server/src/services/alert_service.ts
```diff
@@ -13,7 +13,17 @@ export class Alert {
...
+
+  public recieve(segment: Segment) {
+    if(this.lastSendedSegment === undefined || !segment.equals(this.lastSendedSegment) ) {
+      this.lastSendedSegment = segment;
+      this.sender(this.analyticUnit, segment);
```
rozetko: `sender` doesn't seem like a good name for a function
function name should be a verb (`send`)

server/src/services/alert_service.ts
```diff
@@ -13,7 +13,17 @@ export class Alert {
   public update(now: number) {};
 }
 
-class PatternAlert extends Alert {};
+class PatternAlert extends Alert {
+
+  private lastSendedSegment: Segment;
+
+  public recieve(segment: Segment) {
```
rozetko: `receive`
¯\_(ツ)_/¯

PR #369, opened 2019-01-24
Threshold value is not saving #368

### LINKED ISSUE #368: Threshold value is not saving
Steps to reproduce:

* create threshold analytic unit
* change value
* save dashboard
* refresh

Value would stay default

amper43:
fixes #368

### DIFF
--- a/server/src/models/threshold_model.ts
+++ b/server/src/models/threshold_model.ts
@@ -49,9 +49,7 @@ export class Threshold {
 }
 
 export async function findOne(id: AnalyticUnitId): Promise<Threshold | null> {
-  const query: any = { _id: id };
-
-  const threshold = await db.findOne(query);
+  const threshold = await db.findOne(id);
   if(threshold === null) {
     return null;
   }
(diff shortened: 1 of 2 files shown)


server/src/models/threshold_model.ts
```diff
@@ -49,7 +49,7 @@ export class Threshold {
 }
 
 export async function findOne(id: AnalyticUnitId): Promise<Threshold | null> {
-  const query: any = { id };
+  const query: any = { _id: id };
```
jonyrock:
```suggestion
  const query = { _id: id };
```

jonyrock: you can use https://github.com/hastic/hastic-server/blob/master/server/src/services/data_service.ts#L39

server/src/models/threshold_model.ts
```diff
@@ -61,9 +61,9 @@ export async function findOne(id: AnalyticUnitId): Promise<Threshold | null> {
 export async function updateThreshold(id: AnalyticUnitId, value: number, condition: Condition) {
   const threshold = await db.findOne(id);
   if(threshold === null) {
-    return db.insertOne({ id, value, condition });
+    return db.insertOne({ _id: id, value, condition });
   }
-  return db.updateOne({ id }, { value, condition });
+  return db.updateOne({ _id: id }, { value, condition });
```
jonyrock:
```suggestion
  return db.updateOne(id, { value, condition });
```

PR #371, opened 2019-01-24
Labeled pattern intersects with predicted #253

### LINKED ISSUE #253: Labeled pattern intersects with predicted
On pic you can see which one was selected and detected
image

I think predictor should not detect something which is not labeled.
Btw, if predictor DOES NOT predict it, then it is a BAD predictor. In this case we should write to log that predictor can`t detect something which is labeled.

amper43:
fixes #253

### DIFF
--- a/server/src/models/segment_model.ts
+++ b/server/src/models/segment_model.ts
@@ -96,9 +96,14 @@ export async function findMany(id: AnalyticUnitId, query: FindManyQuery): Promis
 }
 
 export async function insertSegments(segments: Segment[]) {
+  let analyticUnitId: AnalyticUnitId = segments[0].analyticUnitId
   let segmentIdsToRemove: SegmentId[] = [];
   let segmentsToInsert: Segment[] = [];
-  let learningSegments: Segment[] = await db.findMany({labeled: true, deleted: false});
+  let learningSegments: Segment[] = await db.findMany({
+    analyticUnitId: analyticUnitId,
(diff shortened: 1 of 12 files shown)


server/src/models/segment_model.ts
```diff
@@ -96,11 +96,22 @@ export async function findMany(id: AnalyticUnitId, query: FindManyQuery): Promis
 }
 
 export async function insertSegments(segments: Segment[]) {
+  let analyticUnitId: AnalyticUnitId = segments[0].analyticUnitId
```
rozetko:
```suggestion
  let analyticUnitId: AnalyticUnitId = segments[0].analyticUnitId;
```

server/src/models/segment_model.ts
```diff
@@ -96,11 +96,22 @@ export async function findMany(id: AnalyticUnitId, query: FindManyQuery): Promis
...
   let segmentIdsToRemove: SegmentId[] = [];
   let segmentsToInsert: Segment[] = [];
+  let learningSegments: Segment[] = await db.findMany({
+    analyticUnitId,
+    labeled: true,
+    deleted: false
+  });
   for(let segment of segments) {
+    let intersectedLearning = learningSegments.filter(s => segment.from <= s.to && segment.to >= s.from);
```
rozetko: Maybe it would be better to call them `intersectedLabeled`?

PR #396, opened 2019-02-05
No data threshold #370

### LINKED ISSUE #370: No data threshold
Check value for `None`
https://github.com/hastic/hastic-grafana-app/issues/134

rozetko:
Closes #370

### DIFF
--- a/analytics/analytics/detectors/threshold_detector.py
+++ b/analytics/analytics/detectors/threshold_detector.py
@@ -28,15 +28,25 @@ class ThresholdDetector(Detector):
         value = cache['value']
         condition = cache['condition']
 
+        now = int(time()) * 1000
+        segment = ({'from': now, 'to': now})
+        segments = []
+
         dataframe_without_nans = dataframe.dropna()
         if len(dataframe_without_nans) == 0:
-            return dict()
+            if condition == 'No data':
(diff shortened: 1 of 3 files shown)


analytics/analytics/detectors/threshold_detector.py
```diff
@@ -28,15 +28,25 @@ def detect(self, dataframe: pd.DataFrame, cache: Optional[ModelCache]) -> dict:
         value = cache['value']
         condition = cache['condition']
 
+        now = int(time()) * 1000
```
jonyrock: dont like this `... * 1000` thing. Can we make a functoin in utils for this?

server/src/models/threshold_model.ts
```diff
@@ -12,7 +12,8 @@ export enum Condition {
   ABOVE_OR_EQUAL = '>=',
   EQUAL = '=',
   LESS_OR_EQUAL = '<=',
-  LESS = '<'
+  LESS = '<',
+  NO_DATA = 'No data'
```
jonyrock:
```suggestion
  NO_DATA = 'NO_DATA'
```
similar to https://github.com/hastic/hastic-server/blob/master/server/src/models/analytics_task_model.ts#L10

PR #402, opened 2019-02-08
Fix wait for full bucket

amper43:


jonyrock:
@amper43 may I ask you to add a test for this bugfix

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -65,14 +65,18 @@ def detect(self, dataframe: pd.DataFrame, cache: Optional[models.ModelCache]) ->
...
 
+        res = self.detect(self.bucket.data, cache)
         if len(self.bucket.data) >= self.window_size and cache != None:
-            res = self.detect(self.bucket.data, cache)
             excess_data = len(self.bucket.data) - self.max_window_size
             self.bucket.drop_data(excess_data)
-            return res
         
-        return None
+        return res if res else None
```
VargBurz: lets use basic if/else construction, cuz its hard to read..

amper43:
@jonyrock  this is not a bug, just avoiding waiting of full bucket and try run detect with available points instead.

analytics/analytics/detectors/pattern_detector.py
jonyrock:
```suggestion
        if res:
                return res
        else:
                None
```

amper43: @VargBurz  fixed

PR #409, opened 2019-02-12
Config linux source map path overrides

amper43:
just config fixes for linux


jonyrock:
Ok, it is for .vscode debugging

Issue #416, opened 2019-02-14
Accept various URL inputs


sankerust:
Recently there was an issue, where `GRAFANA_URL` provided by the user was not accepted by hastic, since it had a slash (`/`) at the end of the URL.
`"http://localhost:8000/"`
instead of:
`"http://localhost:8000"`

resulted in this error:
`Data kit: got response 404, message: Request failed with status code 404`

We should figure out a way to accept various URL inputs, to avoid such issues in future.

jonyrock:
I think we can also strip whitespaces from the string

PR #429, opened 2019-02-22
Accept various URL inputs #416

sankerust:
closes https://github.com/hastic/hastic-server/issues/416

`GRAFANA_URL` field in config.json is now much more flexible and accepts various inputs.
- no protocol
- no slashes
- slashes at the end of input

Needs tests -> WIP

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -50,9 +50,9 @@
     "ts-jest": "^23.1.1",
     "ts-loader": "^4.4.1",
     "typescript": "^2.8.3",
+    "url-parse": "^1.4.4",
     "webpack": "^4.12.0",
     "webpack-cli": "^3.0.8",
-    "zeromq": "^4.6.0",
-    "url-parse": "^1.4.4"
+    "zeromq": "^4.6.0"
   }
 }
--- a/server/package.json
+++ b/server/package.json
@@ -19,7 +19,9 @@
     "url": "https://github.com/hastic/hastic-server/issues"
   },
   "homepage": "https://github.com/hastic/hastic-server#readme",
-  "dependencies": {},
+  "dependencies": {
+    "url-parse": "^1.4.4"
+  },
   "devDependencies": {
     "@types/jest": "^23.1.1",
     "@types/koa": "^2.0.46",
--- a/server/package.json
+++ b/server/package.json
@@ -19,9 +19,7 @@
     "url": "https://github.com/hastic/hastic-server/issues"
   },
   "homepage": "https://github.com/hastic/hastic-server#readme",
-  "dependencies": {
-    "url-parse": "^1.4.4"
-  },
+  "dependencies": {},
   "devDependencies": {
     "@types/jest": "^23.1.1",
     "@types/koa": "^2.0.46",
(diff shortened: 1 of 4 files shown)


server/package.json
```diff
@@ -19,7 +19,9 @@
     "url": "https://github.com/hastic/hastic-server/issues"
   },
   "homepage": "https://github.com/hastic/hastic-server#readme",
-  "dependencies": {},
+  "dependencies": {
+    "url-parse": "^1.4.4"
```
rozetko: let's move to dev deps

server/src/config.ts
```diff
@@ -58,6 +60,30 @@ function getConfigField(field: string, defaultVal?: any) {
   return val;
 }
 
+function checkUrl(grafanaUrl) {
```
rozetko:
```suggestion
function checkUrl(grafanaUrl: string) {
```

server/src/config.ts
```diff
@@ -58,6 +60,30 @@ function getConfigField(field: string, defaultVal?: any) {
...
+  let urlObj = new url(grafanaUrl);
+  if(urlObj.protocol !== 'http:' && urlObj.protocol !== 'https:') {
+    grafanaUrl = 'http:' + grafanaUrl;
```
rozetko: does `grafanaUrl` have `//`?

server/src/config.ts
```diff
@@ -58,6 +60,30 @@ function getConfigField(field: string, defaultVal?: any) {
...
+    urlObj = new url(grafanaUrl);
+    console.log('No protocol provided in GRAFANA_URL -> inserting "http:"');
+  }
+  if(urlObj.slashes === false) {
+    urlObj = new url(urlObj.protocol + '//' + urlObj.pathname);
```
rozetko:
```suggestion
    urlObj = new url(`${urlObj.protocol}//${urlObj.pathname}`);
```

server/src/config.ts
```diff
@@ -58,6 +60,30 @@ function getConfigField(field: string, defaultVal?: any) {
...
+    console.log('No slashes were provided after the protocol -> inserting slashes');
+  } 
+  if(urlObj.pathname.slice(-1) === '/') {
+    urlObj.pathname = urlObj.pathname.slice(0, -1);
+  }
+  let finalUrl = urlObj.protocol + '//' + urlObj.hostname;
```
rozetko:
```suggestion
  let finalUrl = `${urlObj.protocol}//${urlObj.hostname}`;
```

server/src/config.ts
```diff
@@ -58,6 +60,30 @@ function getConfigField(field: string, defaultVal?: any) {
...
+  if(urlObj.port !== '') {
+    finalUrl = finalUrl + ':' + urlObj.port;
```
rozetko: what if Grafana is at sub-url?
`http://localhost:3000/grafana`
Looks like it would look like `http://localhost/grafana:3000`
I can be wrong

sankerust: Doesn't seem like it:
image

server/src/config.ts
sankerust: Doesn't matter. There is a check for `//` here:
image
If slashes are provided - it is skipped,
if not - it rebuilds the URL.

server/src/config.ts
rozetko: nice

server/src/config.ts
rozetko: Oh, I see

PR #431, opened 2019-02-22
Webhook about missing connection to grafana #412

### LINKED ISSUE #412: System webhook about missing connection to Grafana
We can connect to a datasource through Grafana, but Grafana itself might be unreachable.
For example, missing `API_KEY` can break everything.

amper43:
fixes #412

### DIFF
--- a/server/src/utils/reporter.ts
+++ b/server/src/utils/reporter.ts
@@ -2,7 +2,7 @@ export function availableReporter(
   positiveMsg: string|null,
   negativeMsg: string|null,
   positiveAction = console.log,
-  negativeAction=console.error,
+  negativeAction = console.error,
 ) {
   let reported = false;
   return available => {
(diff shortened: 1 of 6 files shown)


server/src/utils/reporter.ts
```diff
@@ -0,0 +1,23 @@
+export function availableReporter(
+  positiveMsg: string|null,
+  negativeMsg: string|null,
+  positiveAction=console.log,
```
rozetko:
```suggestion
  positiveAction = console.log,
```

server/src/utils/reporter.ts
```diff
@@ -0,0 +1,23 @@
...
+  negativeAction=console.error,
```
rozetko:
```suggestion
  negativeAction = console.error,
```

Issue #264, opened 2018-11-25
Not-ending learning


jonyrock:
I think I am not the only how experienced when you start a "General" learning and it never ends

![image](https://user-images.githubusercontent.com/1022757/48978462-e4ec2800-f0bc-11e8-8a2a-985bbf849e48.png)

jonyrock:
It might be because analytic fails and restarts without restarting learning

jonyrock:
It is not because of restarting. General pattern indeed could hang just with two steps:

* label a big segment
* save
* delete something predicted
* save

rozetko:
```
2018-12-11 12:41:23,076 [Analytics] [ERROR]  handle_analytic_task exception: 'Traceback (most recent call last):
  File "bin/../analytics/analytic_unit_manager.py", line 77, in handle_analytic_task
    result_payload = await self.__handle_analytic_task(task)
  File "bin/../analytics/analytic_unit_manager.py", line 71, in __handle_analytic_task
    return await worker.do_detect(data, payload['cache'])
  File "bin/../analytics/analytic_unit_worker.py", line 34, in do_detect
    return self._detector.detect(data, cache)
  File "bin/../analytics/detectors/pattern_detector.py", line 50, in detect
    detected = self.model.detect(dataframe, cache)
  File "bin/../analytics/models/model.py", line 47, in detect
    ) for x in result]
  File "bin/../analytics/models/model.py", line 47, in <listcomp>
    ) for x in result]
  File "/usr/local/lib/python3.6/site-packages/pandas/core/series.py", line 601, in __getitem__
    result = self.index.get_value(self, key)
  File "/usr/local/lib/python3.6/site-packages/pandas/core/indexes/base.py", line 2477, in get_value
    tz=getattr(series.dtype, 'tz', None))
  File "pandas/_libs/index.pyx", line 98, in pandas._libs.index.IndexEngine.get_value (pandas/_libs/index.c:4404)
  File "pandas/_libs/index.pyx", line 106, in pandas._libs.index.IndexEngine.get_value (pandas/_libs/index.c:4087)
  File "pandas/_libs/index.pyx", line 154, in pandas._libs.index.IndexEngine.get_loc (pandas/_libs/index.c:5126)
  File "pandas/_libs/hashtable_class_helper.pxi", line 759, in pandas._libs.hashtable.Int64HashTable.get_item (pandas/_libs/hashtable.c:14031)
[...]

PR #433, opened 2019-02-25
Futures for detect and recieve

amper43:


rozetko:
Should fix https://github.com/hastic/hastic-server/issues/264

PR #460, opened 2019-03-12
Test models with random values #459

### LINKED ISSUE #459: Test models with random values
Add test with random value generator

VargBurz:
closes #459


analytics/tests/test_dataset.py
```diff
@@ -264,3 +298,9 @@ def create_dataframe(data_val: list) -> pd.DataFrame:
     dataframe = pd.DataFrame(data)
     dataframe['timestamp'] = pd.to_datetime(dataframe['timestamp'], unit='ms')
     return dataframe
+
+def create_random_model(window_size: int) -> list:
+    model = []
```
rozetko: Can be just `return [random.randint(0, 100) for _ in range(window_size * 2 + 1)]`

VargBurz: Is it really easier?

PR #461, opened 2019-03-12
Synchronize data with server #455

### LINKED ISSUE #455: Synchronize data with server #64
Simmetrical issue for https://github.com/hastic/hastic-grafana-app/issues/64

amper43:
fixes #455 
hastic/hastic-grafana-app#64 

Also, splits `panelUrl` to `grafanaUrl` and `panelId` (hastic/hastic-grafana-app#102)


server/src/routes/panel_router.ts
```diff
@@ -11,9 +11,9 @@ async function getAnalyticUnits(ctx: Router.IRouterContext) {
       throw new Error('panelUrl is missing');
     }
 
-    const analyticUnits = await Panel.findOne({ panelUrl });
+    let panel = await Panel.findOne({ panelUrl });
 
-    ctx.response.body = { analyticUnits };
+    ctx.response.body = { analyticUnitViews: panel.analyticUnitViews };
```
rozetko: let's return empty array if `panel === null`

server/src/controllers/analytics_controller.ts
```diff
@@ -284,7 +284,10 @@ export async function runDetect(id: AnalyticUnit.AnalyticUnitId) {
 }
 
 export async function remove(analyticUnitId: AnalyticUnit.AnalyticUnitId) {
-  await cancelAnalyticsTask(analyticUnitId);
+  // We don't await for analytics task cancellation here
+  // If we add await, the rest function would be executed only when analytics becomes up
```
```suggestion
  // If we add await, the rest function will be executed only when analytics becomes up
```

PR #484, opened 2019-03-25
Not-ending learning #264

VargBurz:
#264


analytics/analytics/detectors/pattern_detector.py
```diff
@@ -66,21 +66,24 @@ def detect(self, dataframe: pd.DataFrame, cache: Optional[models.ModelCache]) ->
...
     def recieve_data(self, data: pd.DataFrame, cache: Optional[ModelCache]) -> Optional[dict]:
+        logging.debug('Start recieve_data')
         data_without_nan = data.dropna()
 
         if len(data_without_nan) == 0:
             return None
 
         self.bucket.receive_data(data_without_nan)
-        if not cache: cache = {}
+        if not cache:
```
rozetko: cache == None?

amper43: @rozetko `cache` may be `{}`

analytics/analytics/analytic_unit_manager.py
```diff
@@ -64,7 +64,7 @@ def __ensure_worker(
             returns payload or None
         """
         analytic_unit_id: AnalyticUnitId = task['analyticUnitId']
-
+        log.debug('Analytics get task with type: {}'.format(task['type']))
```
amper43: we already log task types in top level code

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -66,21 +66,24 @@ def detect(self, dataframe: pd.DataFrame, cache: Optional[models.ModelCache]) ->
         }
 
     def recieve_data(self, data: pd.DataFrame, cache: Optional[ModelCache]) -> Optional[dict]:
+        logging.debug('Start recieve_data')
```
amper43: we already log `PUSH` messages that run `recieve_data`

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -66,21 +66,24 @@ def detect(self, dataframe: pd.DataFrame, cache: Optional[models.ModelCache]) ->
...
         data_without_nan = data.dropna()
 
         if len(data_without_nan) == 0:
             return None
 
         self.bucket.receive_data(data_without_nan)
-        if not cache: cache = {}
+        if not cache:
+            logging.warning('Cache in receive with some problem: {}'.format(cache))
```
amper43: empty `cache` is not a problem, just log `recieve empty cache for task <task id>` or somethink like that

analytics/analytics/models/general_model.py
```diff
@@ -43,6 +43,7 @@ def find_segment_center(self, dataframe: pd.DataFrame, start: int, end: int) ->
     def do_fit(self, dataframe: pd.DataFrame, labeled_segments: list, deleted_segments: list, learning_info: dict) -> None:
         data = utils.cut_dataframe(dataframe)
         data = data['value']
+        logging.debug('Start method do_fit')
```
amper43: if we create 100 analytic units, we got 100 messages `Start method do_fit`. So, without analytic unit id we can't resolve which task wrote this message.

analytics/analytics/models/general_model.py
```diff
@@ -61,6 +62,7 @@ def do_fit(self, dataframe: pd.DataFrame, labeled_segments: list, deleted_segmen
 
         self.state['convolve_min'], self.state['convolve_max'] = utils.get_min_max(convolve_list, self.state['WINDOW_SIZE'] / 3)
         self.state['conv_del_min'], self.state['conv_del_max'] = utils.get_min_max(del_conv_list, self.state['WINDOW_SIZE'])
+        logging.debug('Method do_fit completed correctly')
```
amper43: please, add task or analytic unit ids, in this way we could trace task execution

analytics/analytics/models/general_model.py
```diff
@@ -69,36 +71,40 @@ def do_detect(self, dataframe: pd.DataFrame) -> list:
...
                 delete_list.append(val)
                 continue
-            if (self.all_conv[val] < self.state['conv_del_max'] * 1.02 and self.all_conv[val] > self.state['conv_del_min'] * 0.98):
+            if (watch_conv < self.state['conv_del_max'] * 1.02 and watch_conv > self.state['conv_del_min'] * 0.98):
 	            delete_list.append(val)
 
         for item in delete_list:
             segments.remove(item)
-
+        logging.debug('Filtering ended with pattern indexes {}'.format(set(segments)))
```
amper43: Will it informative with 30+ segments? We already log number of segments found in detection.

VargBurz: its ok. simpler to track filtering

analytics/analytics/analytic_unit_manager.py
VargBurz: want to see all steps for simpler debug

analytics/analytics/detectors/pattern_detector.py
jonyrock: we differentiate `None` and `{}` so it should be different checks

analytics/analytics/models/general_model.py
```diff
@@ -61,44 +61,41 @@ def do_fit(self, dataframe: pd.DataFrame, labeled_segments: list, deleted_segmen
 
         self.state['convolve_min'], self.state['convolve_max'] = utils.get_min_max(convolve_list, self.state['WINDOW_SIZE'] / 3)
         self.state['conv_del_min'], self.state['conv_del_max'] = utils.get_min_max(del_conv_list, self.state['WINDOW_SIZE'])
+        logging.debug('Method do_fit completed correctly for analytic unit: {}'.format(id))
 
-    def do_detect(self, dataframe: pd.DataFrame) -> list:
+    def do_detect(self, dataframe: pd.DataFrame, id: str) -> list:
```
amper43: Which type have list's elements?

analytics/analytics/models/general_model.py
```diff
@@ -61,44 +61,41 @@ def do_fit(self, dataframe: pd.DataFrame, labeled_segments: list, deleted_segmen
...
+        all_corr_peaks = utils.find_peaks(all_corr, window_size * 2)
         filtered = self.__filter_detection(all_corr_peaks, data)
+        filtered = list(filtered)
+        logging.debug('Method do_detect completed correctly for analytic unit: {}'.format(id))
         return set(item + window_size for item in filtered)
 
-    def __filter_detection(self, segments: list, data: list):
-        if len(segments) == 0 or len(self.state.get('pattern_center', [])) == 0:
+    def __filter_detection(self, segments:  Generator[int, None, None], data: pd.Series) -> Generator[int, None, list]:
+        if len(self.state.get('pattern_center', [])) == 0:
```
amper43: just `if not state.get('pattern_center')` like in pep008:

```
For sequences, (strings, lists, tuples), use the fact that empty sequences are false.

Yes: if not seq:
     if seq:

No:  if len(seq):
     if not len(seq):
```

analytics/analytics/models/general_model.py
```diff
@@ -61,44 +61,41 @@ def do_fit(self, dataframe: pd.DataFrame, labeled_segments: list, deleted_segmen
...
-
-        for item in delete_list:
-            segments.remove(item)
-
-        return set(segments)
+            if val < PEARSON_COEFF:
+                continue
+            if (watch_conv < self.state['conv_del_max'] * 1.02 and watch_conv > self.state['conv_del_min'] * 0.98):
+                continue
+            yield ind
```
amper43: if nothink will be yielded, method return `None` instead of `List[int]`

analytics/analytics/models/general_model.py
```diff
@@ -61,44 +61,41 @@ def do_fit(self, dataframe: pd.DataFrame, labeled_segments: list, deleted_segmen
...
+        all_corr = utils.create_correlation_data(data, window_size, pat_data)
+        all_corr_peaks = utils.find_peaks(all_corr, window_size * 2)
         filtered = self.__filter_detection(all_corr_peaks, data)
+        filtered = list(filtered)
+        logging.debug('Method do_detect completed correctly for analytic unit: {}'.format(id))
         return set(item + window_size for item in filtered)
 
-    def __filter_detection(self, segments: list, data: list):
-        if len(segments) == 0 or len(self.state.get('pattern_center', [])) == 0:
+    def __filter_detection(self, segments:  Generator[int, None, None], data: pd.Series) -> Generator[int, None, list]:
```
amper43: may be `-> Generator[int, None, List[int]]` ?

analytics/analytics/utils/common.py
```diff
@@ -72,12 +75,14 @@ def timestamp_to_index(dataframe, timestamp):
...
     return time_ind
 
-def peak_finder(data, size):
-    all_max = []
-    for i in range(size, len(data) - size):
-        if data[i] == max(data[i - size: i + size]) and data[i] > data[i + 1]:
-            all_max.append(i)
-    return all_max
+def find_peaks(data: Generator[float, None, None], size: int) -> Generator[float, None, None]:
+    window = deque(islice(data, size*2 + 1))
```
amper43:
```suggestion
    window = deque(islice(data, size * 2 + 1))
```

analytics/analytics/utils/common.py
```diff
@@ -73,12 +75,14 @@ def timestamp_to_index(dataframe, timestamp):
...
+    for i, v in enumerate(data, size):
+        current = window[size]
+        if current == max(window) and current != window[size + 1]:
```
jonyrock: bad

analytics/analytics/utils/common.py
```diff
@@ -224,20 +228,12 @@ def get_convolve(segments: list, av_model: list, data: pd.Series, window_size: i
             convolve_list.append(max(convolve_segment))
     return convolve_list
 
-def create_correlation_data(data: pd.Series, window_size: int, pattern_model: list) -> list:
-    all_corr = []
-    timeout = 120
-    start_time = time.time()
+def create_correlation_data(data: pd.Series, window_size: int, pattern_model: List[float]) -> Generator[float, None, None]:
```
jonyrock: looks weird

analytics/analytics/models/general_model.py
VargBurz: changed to `List[int]`

analytics/analytics/models/general_model.py
jonyrock: we don't use this PEP. We do everything explicit.

jonyrock: I don't like `if len(self.state.get('pattern_center', []))`
cuz in `if` you have too complected expression here

analytics/analytics/utils/common.py
```diff
@@ -72,12 +74,15 @@ def timestamp_to_index(dataframe, timestamp):
...
-    all_max = []
-    for i in range(size, len(data) - size):
-        if data[i] == max(data[i - size: i + size]) and data[i] > data[i + 1]:
-            all_max.append(i)
-    return all_max
+def find_peaks(data: Generator[float, None, None], size: int) -> Generator[float, None, None]:
+    window = deque(islice(data, size * 2 + 1))
+    for i, v in enumerate(data, size):
+        current = window[size]
+        #TODO: remove max() from cycle
```
```suggestion
        #TODO: remove max() from loop
```

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -73,8 +73,8 @@ def recieve_data(self, data: pd.DataFrame, cache: Optional[ModelCache]) -> Optio
             return None
 
         self.bucket.receive_data(data_without_nan)
-        if not cache:
-            logging.debug('Recieve empty cache for task {}'.format(self.analytic_unit_id))
+        if cache == None:
```
jonyrock: > Comparisons to singletons like None should always be done with `is` or `is not`, never the equality operators.

https://www.python.org/dev/peps/pep-0008/

@amper43 @VargBurz 
we do check everything explicit, but when we do it explicit, we do it right

analytics/analytics/models/general_model.py
```diff
@@ -40,7 +39,8 @@ def find_segment_center(self, dataframe: pd.DataFrame, start: int, end: int) ->
         center_ind = start + math.ceil((end - start) / 2)
         return center_ind
 
-    def do_fit(self, dataframe: pd.DataFrame, labeled_segments: list, deleted_segments: list, learning_info: dict) -> None:
+    def do_fit(self, dataframe: pd.DataFrame, labeled_segments: list, deleted_segments: list, learning_info: dict, AnalyticUnitId: str) -> None:
```
jonyrock: wtf, guys? `AnalyticUnitId` as name of argument?
@VargBurz @rozetko @amper43

rozetko: Oops, didn't notice

Issue #481, opened 2019-03-25
Timeouts for fit and detect


rozetko:
We should add timeouts for fit and detect execution and throw exceptions
Maybe they should be configurable

jonyrock:
... ad return proper errors

jonyrock:
@rozetko what is the status of this issue?

rozetko:
We've added timeout for learning but not for detection

PR #485, opened 2019-03-25
Timeout for train #481

rozetko:
#481 
Couldn't add timeout for detect yet because it runs outside executor
image

### DIFF
--- a/analytics/analytics/analytic_unit_worker.py
+++ b/analytics/analytics/analytic_unit_worker.py
@@ -17,17 +17,17 @@ class AnalyticUnitWorker:
         self.analytic_unit_id = analytic_unit_id
         self._detector = detector
         self._executor: Executor = executor
-        self._training_feature: asyncio.Future = None
+        self._training_future: asyncio.Future = None
 
     async def do_train(
         self, payload: Union[list, dict], data: pd.DataFrame, cache: Optional[ModelCache]
     ) -> ModelCache:
-        self._training_feature = self._executor.submit(
+        self._training_future = self._executor.submit(
             self._detector.train, data, payload, cache
         )
         try:
             # TODO: configurable timeout
-            new_cache: ModelCache = self._training_feature.result(timeout = TRAIN_TIMEOUT)
+            new_cache: ModelCache = self._training_future.result(timeout = TRAIN_TIMEOUT)
             return new_cache
         except CancelledError as e:
             return cache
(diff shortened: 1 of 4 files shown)


analytics/analytics/analytic_unit_worker.py
```diff
@@ -4,11 +4,12 @@
 import pandas as pd
 from typing import Optional, Union
 from models import ModelCache
-from concurrent.futures import Executor, CancelledError
+from concurrent.futures import Executor, CancelledError, TimeoutError
 import asyncio
 
 logger = logging.getLogger('AnalyticUnitWorker')
 
+TRAIN_TIMEOUT = 120
```
jonyrock: please add a comment "120 # seconds " or something like this

analytics/analytics/analytic_unit_worker.py
```diff
@@ -21,14 +22,17 @@ def __init__(self, analytic_unit_id: str, detector: detectors.Detector, executor
...
         self, payload: Union[list, dict], data: pd.DataFrame, cache: Optional[ModelCache]
     ) -> ModelCache:
-        self._training_feature = asyncio.get_event_loop().run_in_executor(
-            self._executor, self._detector.train, data, payload, cache
+        self._training_feature = self._executor.submit(
+            self._detector.train, data, payload, cache
         )
         try:
-            new_cache: ModelCache = await self._training_feature
+            # TODO: configurable timeout
```
jonyrock: is it still in "TODO" ?

analytics/analytics/analytic_unit_worker.py
```diff
@@ -21,14 +22,17 @@ def __init__(self, analytic_unit_id: str, detector: detectors.Detector, executor
     async def do_train(
         self, payload: Union[list, dict], data: pd.DataFrame, cache: Optional[ModelCache]
     ) -> ModelCache:
-        self._training_feature = asyncio.get_event_loop().run_in_executor(
-            self._executor, self._detector.train, data, payload, cache
+        self._training_feature = self._executor.submit(
```
jonyrock: is it `training_feature` or `trainint_future`
https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.submit

analytics/analytics/analytic_unit_worker.py
```diff
@@ -21,14 +22,17 @@ def __init__(self, analytic_unit_id: str, detector: detectors.Detector, executor
...
+            self._detector.train, data, payload, cache
         )
         try:
-            new_cache: ModelCache = await self._training_feature
+            # TODO: configurable timeout
+            new_cache: ModelCache = self._training_feature.result(timeout = TRAIN_TIMEOUT)
             return new_cache
         except CancelledError as e:
             return cache
```
jonyrock: it's strange that we return `cache` there. Shouldn't we `raise` anything there?

jonyrock: but maybe fix it in this PR maybe be a bad idea

Issue #489, opened 2019-03-29
Send data to detection in chunks


rozetko:
Detection works synchronously and blocks event loop but we run detection on the whole dataset which may take a while
It would be better to split dataset by chunks and run detect on them separately

jonyrock:
Maybe we can use async generators for this

rozetko:
Closed via https://github.com/hastic/hastic-server/pull/496

PR #496, opened 2019-04-01
Send data to detection in chunks #489 (WIP)

amper43:


analytics/analytics/detectors/pattern_detector.py
```diff
@@ -33,11 +34,13 @@ def resolve_model_by_pattern(pattern: str) -> models.Model:
 AnalyticUnitId = str
 class PatternDetector(Detector):
 
+    MIN_BUCKET_SIZE = 150
+    CHUNK_SIZE = 1500
```
rozetko: why do we need constant chink size?

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -52,9 +55,16 @@ def train(self, dataframe: pd.DataFrame, segments: list, cache: Optional[models.
...
 
-        segments = [{ 'from': segment[0], 'to': segment[1] } for segment in detected['segments']]
+        chunks = self.__get_data_chunks(dataframe)
+
+        segments = []
+        segment_parser = lambda segment: { 'from': segment[0], 'to': segment[1] }
+        for chunk in chunks:
+            detected = self.model.detect(chunk, cache)
+            segments.extend(map(segment_parser, detected['segments']))
+            asyncio.sleep(0)
```
rozetko:
```suggestion
            await asyncio.sleep(0)
```

rozetko: Also, I've heard that it's the best practice to execute `asyncio.sleep` in the begging of the loop
But it's better to verify that info

analytics/analytics/detectors/pattern_detector.py
amper43: it will be changed

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -52,9 +55,16 @@ def train(self, dataframe: pd.DataFrame, segments: list, cache: Optional[models.
     def detect(self, dataframe: pd.DataFrame, cache: Optional[models.ModelCache]) -> dict:
         logger.debug('Unit {} got {} data points for detection'.format(self.analytic_unit_id, len(dataframe)))
         # TODO: split and sleep (https://github.com/hastic/hastic-server/pull/124#discussion_r214085643)
-        detected = self.model.detect(dataframe, self.analytic_unit_id, cache)
 
-        segments = [{ 'from': segment[0], 'to': segment[1] } for segment in detected['segments']]
+        chunks = self.__get_data_chunks(dataframe)
+
+        segments = []
```
rozetko: Maybe it would be better to make it `set` to avoid segments duplication?

analytics/analytics/detectors/pattern_detector.py
rozetko: It should depend on analytic unit's window size

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -33,11 +34,13 @@ def resolve_model_by_pattern(pattern: str) -> models.Model:
 AnalyticUnitId = str
 class PatternDetector(Detector):
 
+    MIN_BUCKET_SIZE = 150
```
rozetko: It's strange that you define this constant but don't use it

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -88,3 +98,11 @@ def recieve_data(self, data: pd.DataFrame, cache: Optional[ModelCache]) -> Optio
             return res
         else:
             return None
+
+    def __get_data_chunks(self, dataframe: pd.DataFrame) -> Generator[pd.DataFrame, None, None]:
+        data_len = len(dataframe)
+        end_of_slice = lambda offset: offset + min(self.CHUNK_SIZE, data_len - offset)
+        chunk_slicer = lambda offset: slice(offset, end_of_slice(offset))
+        return (dataframe[chunk_slicer(offset)]
+                    for offset in range(0, data_len, self.CHUNK_SIZE)
```
rozetko: Chunks don't seem to be intersected but they should be
Otherwise, we could miss segments

analytics/analytics/analytic_unit_worker.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, analytic_unit_id: str, detector: detectors.Detector, executor
             raise Exception('Timeout ({}s) exceeded while learning'.format(config.LEARNING_TIMEOUT))
 
     async def do_detect(self, data: pd.DataFrame, cache: Optional[ModelCache]) -> dict:
-        return self._detector.detect(data, cache)
+        return await self._detector.detect(data, cache)
```
rozetko: It would be better `await`ed higher (e.g. in manager)

analytics/analytics/detectors/pattern_detector.py
rozetko: Seems like you forgot to delete it

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -49,12 +52,20 @@ def train(self, dataframe: pd.DataFrame, segments: list, cache: Optional[models.
...
         }
 
-    def detect(self, dataframe: pd.DataFrame, cache: Optional[models.ModelCache]) -> dict:
+    async def detect(self, dataframe: pd.DataFrame, cache: Optional[models.ModelCache]) -> dict:
         logger.debug('Unit {} got {} data points for detection'.format(self.analytic_unit_id, len(dataframe)))
         # TODO: split and sleep (https://github.com/hastic/hastic-server/pull/124#discussion_r214085643)
-        detected = self.model.detect(dataframe, self.analytic_unit_id, cache)
 
-        segments = [{ 'from': segment[0], 'to': segment[1] } for segment in detected['segments']]
+        window_size = cache.get('WINDOW_SIZE', 0)
```
rozetko: I guess it would be better to throw an exception if `window_size` is undefined

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -88,3 +110,31 @@ def recieve_data(self, data: pd.DataFrame, cache: Optional[ModelCache]) -> Optio
...
+            nonintersected = chunk_size - intersection
+            mod = data_len % nonintersected
+            chunks_number = data_len // nonintersected
+
+            offset = 0
+            for i in range(chunks_number):
+                yield slice(offset, offset + nonintersected + 1)
+                offset += nonintersected
+
+            yield slice(offset, offset + mod)
```
jonyrock: is it correct to return `[]` here sometimes ?

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -88,3 +110,31 @@ def recieve_data(self, data: pd.DataFrame, cache: Optional[ModelCache]) -> Optio
...
         else:
             return None
+
+    def __get_data_chunks(self, dataframe: pd.DataFrame, window_size: int) -> Generator[pd.DataFrame, None, None]:
+        """
+        TODO: fix description
+        Return generator, that yields dataframe's chunks. Chunks have 3 WINDOW_SIZE length and 2 WINDOW_SIZE step.
+        Example: recieved dataframe: [0, 1, 2, 3, 4, 5], returned chunks [0, 1, 2], [2, 3, 4], [4, 5].
+        """
+        chunk_size = window_size * 100
```
jonyrock: what does this `100` means? How is it connected with `3 WINDOW_SIZE`

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -88,3 +110,31 @@ def recieve_data(self, data: pd.DataFrame, cache: Optional[ModelCache]) -> Optio
...
+        intersection = window_size
+
+        data_len = len(dataframe)
+
+        if data_len < chunk_size:
+            return (chunk for chunk in (dataframe,))
```
jonyrock: I think you can just do `dataframe`  and then `return`

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -76,7 +98,7 @@ def recieve_data(self, data: pd.DataFrame, cache: Optional[ModelCache]) -> Optio
         if cache == None:
             logging.debug('Recieve_data cache is None for task {}'.format(self.analytic_unit_id))
             cache = {}
-        bucket_size = max(cache.get('WINDOW_SIZE', 0) * 3, self.min_bucket_size)
+        bucket_size = max(cache.get('WINDOW_SIZE', 0) * 3, self.MIN_BUCKET_SIZE)
```
jonyrock: is in this terminology `bucket` and `chunk` are the same thing ?

PR #502, opened 2019-04-02
Batch detection #500

### LINKED ISSUE #500: Batch detection
Subtask for #497

rozetko:
Fixes #500

### DIFF
--- a/server/src/routes/analytic_units_router.ts
+++ b/server/src/routes/analytic_units_router.ts
@@ -125,7 +125,8 @@ async function deleteUnit(ctx: Router.IRouterContext) {
 async function runDetect(ctx: Router.IRouterContext) {
   const { ids } = ctx.request.body as { ids: AnalyticUnit.AnalyticUnitId[] };
 
-  _.each(ids, async id => await AnalyticsController.runFirstLearning(id));
+  await Promise.all(ids.map(AnalyticsController.runFirstLearning));
+
   ctx.response.body = {
     code: 200,
     message: 'Success'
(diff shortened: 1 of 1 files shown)


server/src/routes/analytic_units_router.ts
```diff
@@ -122,8 +123,9 @@ async function deleteUnit(ctx: Router.IRouterContext) {
 }
 
 async function runDetect(ctx: Router.IRouterContext) {
-  const { id: analyticUnitId } = ctx.request.body as { id: AnalyticUnit.AnalyticUnitId };
```
jonyrock: I thing we can take both `id` and `ids` 
what do you think

rozetko: Not sure if it would be useful
`ids` is more universal

server/src/routes/analytic_units_router.ts
```diff
@@ -122,8 +123,9 @@ async function deleteUnit(ctx: Router.IRouterContext) {
...
-  AnalyticsController.runFirstLearning(analyticUnitId);
+  const { ids } = ctx.request.body as { ids: AnalyticUnit.AnalyticUnitId[] };
+
+  _.each(ids, async id => await AnalyticsController.runFirstLearning(id));
```
jonyrock: We don't you do `Promise.all(ids.map(AnalyticsController.runFirstLearning))` here?

server/src/routes/analytic_units_router.ts
jonyrock: ok

PR #507, opened 2019-04-04
Find start and end of peaks and troughs #506

### LINKED ISSUE #506: Find start and end of peaks and troughs
We have only patterns center to determine the exact location in dataset.
We should look for start and end of patterns.

VargBurz:
Closes #506 

Problem: Location of found patterns depended only on their centers. That's why it was not clear where the pattern really starts and ends.
Solution: Use confidence to determine the exact start and end of a pattern. Confidence is obtained from learning.
For peaks: we assume that everything above confidence is the peak. Next we go through right and left parts of segment. The pattern ends when the trend changes.
For troughs: we convert trough to peak by subtracting the maximum element from each value. Then the peak algorithm is used.

Also, we assume that the `window_size` of values is the maximum border of the pattern from the center.

Other changes:
- `CONFIDENCE_FACTOR` is increased from 0.2 to 0.5 (to minimize false detections)


analytics/analytics/utils/common.py
```diff
@@ -191,7 +191,7 @@ def find_extremum_index(segment: np.ndarray, selector: bool) -> int:
     else:
         return segment.argmin()
 
-def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
+def get_interval(data: pd.Series, center: int, window_size: int, normalization = False) -> pd.Series:
```
rozetko: Let's add a docstring for this method

analytics/analytics/utils/common.py
```diff
@@ -201,7 +201,48 @@ def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
...
         right_bound = len(data)
-    return data[left_bound: right_bound]
+    result_interval = data[left_bound: right_bound]
+    if normalization:
+        result_interval = subtract_min_without_nan(result_interval)
+    return result_interval
+
+def get_borders_of_peak(pattern_center: List[int], data: pd.Series, window_size: int, confidence: float, reverse=False) -> List[int]:
+    #Find start and end of patterns for peak
+    #trough will be convert in peak
```
```suggestion
    #if reverse == True - data will be inversed (trough -> peak / peak -> trough)
```

analytics/analytics/utils/common.py
```diff
@@ -201,7 +201,48 @@ def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
...
     if right_bound > len(data):
         right_bound = len(data)
-    return data[left_bound: right_bound]
+    result_interval = data[left_bound: right_bound]
+    if normalization:
+        result_interval = subtract_min_without_nan(result_interval)
+    return result_interval
+
+def get_borders_of_peak(pattern_center: List[int], data: pd.Series, window_size: int, confidence: float, reverse=False) -> List[int]:
+    #Find start and end of patterns for peak
```
```suggestion
    #Find start and end of pattern for peak
```

analytics/analytics/utils/common.py
```diff
@@ -201,7 +201,48 @@ def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
         left_bound = 0
     if right_bound > len(data):
         right_bound = len(data)
-    return data[left_bound: right_bound]
+    result_interval = data[left_bound: right_bound]
+    if normalization:
+        result_interval = subtract_min_without_nan(result_interval)
+    return result_interval
+
+def get_borders_of_peak(pattern_center: List[int], data: pd.Series, window_size: int, confidence: float, reverse=False) -> List[int]:
```
```suggestion
def get_borders_of_peaks(pattern_centers: List[int], data: pd.Series, window_size: int, confidence: float, reverse=False) -> List[int]:
```

analytics/analytics/utils/common.py
```diff
@@ -201,7 +201,48 @@ def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
...
+        left_segment = current_pattern[:window_size]
+        right_segment = current_pattern[window_size:]
+        left_border = get_end_of_segment(left_segment.iloc[::-1])
+        right_border = get_end_of_segment(right_segment)
+        border_list.append((left_border, right_border))
+    return border_list
+
+def get_end_of_segment(segment: pd.Series, positive = True) -> int:
+    #Find end of descending part of pattern
+    #zone of admissibility: allowable error - 1 index
```
```suggestion
    #allowable error is 1 index
```

analytics/analytics/utils/common.py
```diff
@@ -201,7 +201,48 @@ def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
...
+    if len(segment) == 0:
+        return 1
+    for ind in range(1, len(segment) - 1):
```
rozetko: `ind` -> `idx`

analytics/analytics/utils/common.py
```diff
@@ -201,7 +201,48 @@ def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
...
+            current_pattern = reverse_segment(current_pattern)
+        current_pattern = current_pattern - confidence
+        left_segment = current_pattern[:window_size]
+        right_segment = current_pattern[window_size:]
+        left_border = get_end_of_segment(left_segment.iloc[::-1])
+        right_border = get_end_of_segment(right_segment)
+        border_list.append((left_border, right_border))
+    return border_list
+
+def get_end_of_segment(segment: pd.Series, positive = True) -> int:
```
rozetko: Please add the meaning of `positive` argument to docstring

analytics/analytics/utils/common.py
```diff
@@ -191,7 +191,11 @@ def find_extremum_index(segment: np.ndarray, selector: bool) -> int:
     else:
         return segment.argmin()
 
-def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
+def get_interval(data: pd.Series, center: int, window_size: int, normalization = False) -> pd.Series:
+    """
+    Get an interval window_size to the left and right of center
```
```suggestion
    Get an interval with 2*window_size length
    window_size to the left, window_size to the right of center
```

analytics/analytics/utils/common.py
```diff
@@ -201,7 +205,59 @@ def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
...
+        left_border = get_end_of_segment(left_segment, descending = False)
+        right_border = get_end_of_segment(right_segment)
+        border_list.append((left_border, right_border))
+    return border_list
+
+def get_end_of_segment(segment: pd.Series, positive = True, descending = True) -> int:
+    """
+    Find end of descending or ascending part of pattern
+    Allowable error is 1 index
+    If positive == True - skip positive values
```
rozetko: Let's rename `positive` to `skip_positive_items`

```suggestion
    If positive == True - skip positive values 
    If descending == True, check if values are descending, else - ascending
```

analytics/analytics/utils/common.py
```diff
@@ -201,7 +206,60 @@ def get_interval(data: pd.Series, center: int, window_size: int) -> pd.Series:
...
+        right_border = get_end_of_segment(right_segment)
+        border_list.append((left_border, right_border))
+    return border_list
+
+def get_end_of_segment(segment: pd.Series, positive = True, descending = True) -> int:
+    """
+    Find end of descending or ascending part of pattern
+    Allowable error is 1 index
+    If positive == True - skip positive values 
+    If descending == True - check if values are descending, else - ascending
```
```suggestion
```

PR #515, opened 2019-04-05
add AsyncZmqThread and AsyncZmqActor classes

jonyrock:
We need some multi threading code. The best way to do it is to use Actor Model. 

This PR contains an implementation of Actor Model based on zmq, asyncio and threads
We haven't choose existing Actor implementation cuz we want zmq. 

## Implementation details

* zqm should handle queue from/to actor. So we should always be able to await next message from actor

* Inheritance is: `Thread <- AsyncZmqThread <- AsyncZmqActor` so it's possible to do `myActor.join()` but please don't do it.


analytics/analytics/utils/concurrent.py
```diff
@@ -0,0 +1,100 @@
...
+    async def put_message(self, message: str):
+        await self.__actor_socket.send_string(message)
+    
+    def __aiter__(self):
+        return self
+    
+    async def __anext__(self) -> str:
+        return await self.__actor_socket.recv_string()
+    
+    # TODO: implement graceful stopping
```
jonyrock:
```suggestion
    # TODO: implement graceful stopping

```

analytics/analytics/utils/concurrent.py
```diff
@@ -0,0 +1,100 @@
...
+    async def __message_loop(self):
+        while True:
+            text = await self.__zmq_socket.recv_string()
+            asyncio.ensure_future(self._on_message(text))
+
+    async def _send_message(self, message: str):
+        await self.__zmq_socket.send_string(message)
+
+    async def _on_message(self, message: str):
+        """Override this method to receive messages"""
```
rozetko: We can wrap it in `@abc.abstractmethod` decorator
https://docs.python.org/3.6/library/abc.html#abc.abstractmethod

analytics/analytics/utils/concurrent.py
```diff
@@ -0,0 +1,100 @@
...
+    * async _run(self) -- required
+    * async def _on_message(self, message: str) -- not required
+
+    then run asyncZmqActor.start()
+
+    Example:
+
+    >>>
+    class MyActor(AsyncZmqActor):
+        def async _run(self):
```
rozetko: `def async` -> `async def`

analytics/analytics/utils/concurrent.py
```diff
@@ -0,0 +1,100 @@
...
+        self.__asyncio_loop.run_until_complete(self._run())
+
+    # TODO: implement stop signal handling
+
+
+class AsyncZmqActor(AsyncZmqThread):
+    """Threaded and Async Actor model based on ZMQ inproc communication
+
+    override following
+    * async _run(self) -- required
```
amper43:
```suggestion
    * async def _run(self) -- required
```

analytics/analytics/utils/concurrent.py
```diff
@@ -0,0 +1,100 @@
...
+    * async def _run(self) -- required
+    * async def _on_message(self, message: str) -- not required
+
+    then run asyncZmqActor.start()
+
+    Example:
+
+    >>>
+    class MyActor(AsyncZmqActor):
+        def async _run(self):
```
jonyrock:
```suggestion
        async def _run(self):
```

PR #522, opened 2019-04-06
AsyncZmqThread and AsyncZmqActor better method names and docs

jonyrock:
It's difficult to understand destinations and origins of messages when you implement your Actor. 
So in this PR method names become a bit more explicit, docs are also reviewed. 

`AsyncZmqActor` is not async iterator now.

### DIFF
--- a/analytics/analytics/utils/concurrent.py
+++ b/analytics/analytics/utils/concurrent.py
@@ -5,7 +5,9 @@ import zmq.asyncio
 from abc import ABC, abstractmethod
 
 
+# This const defines Thread <-> Actor zmq one-to-one connection
 # We create a seperate zmq context, so zqm address 'inproc://xxx' doesn't matter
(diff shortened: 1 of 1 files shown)


analytics/analytics/utils/concurrent.py
```diff
@@ -5,35 +5,41 @@
 from abc import ABC, abstractmethod
```
amper43: may be it have sense to add comment about import order?

jonyrock: There is no problem with import model actually, but it's worth to mention about MROs (method resolution order) problem in `class AsyncZmqThread(threading.Thread, ABC)` declaration. It's indeed leads to exception if I change to `class AsyncZmqThread(ABC, threading.Thread)`

I saw that error on the first line because VSCode showed it there.

PR #524, opened 2019-04-06
Analytic units' panel fields are not saved to db #523

### LINKED ISSUE #523: Analytic units' panel fields are not saved to db
Fields used only by panel (e.g. `labeledColor`, `detectorType`, etc) are not saved to db
Affects:
https://github.com/hastic/hastic-grafana-app/issues/233
https://github.com/hastic/hastic-grafana-app/issues/232

rozetko:
Fixes #523 

Changes:
- Make `toObject()` `AnalyticUnit`'s method return all the fields

### DIFF
--- a/server/src/controllers/analytics_controller.ts
+++ b/server/src/controllers/analytics_controller.ts
@@ -199,6 +199,7 @@ export async function runLearning(id: AnalyticUnit.AnalyticUnitId) {
       await AnalyticUnitCache.create(id);
     }
 
+    // TODO: create analytics serialization method in AnalyticUnit
     let analyticUnitType = analyticUnit.type;
     let detector = AnalyticUnit.getDetectorByType(analyticUnitType);
     let taskPayload: any = { detector, analyticUnitType, cache: oldCache };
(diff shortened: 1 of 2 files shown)


server/src/models/analytic_unit_model.ts
```diff
@@ -115,7 +115,11 @@ export class AnalyticUnit {
       alert: this.alert,
       lastDetectionTime: this.lastDetectionTime,
       status: this.status,
-      error: this.error
+      error: this.error,
+      labeledColor: this.labeledColor,
```
jonyrock: can we use "positiveLabelColor"

server/src/models/analytic_unit_model.ts
```diff
@@ -115,7 +115,11 @@ export class AnalyticUnit {
...
+      deletedColor: this.deletedColor,
```
jonyrock: and "negativeLabelColor"

server/src/controllers/analytics_controller.ts
```diff
@@ -199,6 +199,7 @@ export async function runLearning(id: AnalyticUnit.AnalyticUnitId) {
       await AnalyticUnitCache.create(id);
     }
 
+    // TODO: create analytics serialization method in AnalyticUnit
```
jonyrock:
```suggestion
    // TODO: create an analytics serialization method in AnalyticUnit
```

server/src/models/analytic_unit_model.ts
rozetko: Let's do it in the separate PR cause it involves DB migration

Issue #316, opened 2018-12-11
Analytics is down while learning / detection


rozetko:
Node doesn't get `ping` response from analytics while learning / detection
Thus, you get `Analytics is down` message
Though it is not down
![image](https://user-images.githubusercontent.com/1989898/49802185-41a63e80-fd5d-11e8-88ed-08c409225d7f.png)

Issue #407, opened 2019-02-11
Analytics gets down without any exception


rozetko:
Sometimes analytics is down but you can't see exception in logs
It happens during detection for webhooks
Maybe executor thread falls and exception just not logged

rozetko:
P.S. that fix didn't help #400

amper43:
tried to fix this issue in https://github.com/hastic/hastic-server/commit/6b9a4a03cc845b53cdcf5ecf9e501cba0f28f4fd

jonyrock:
@rozetko I close this one

PR #527, opened 2019-04-06
Server service actor

jonyrock:
## Motivation

There is still a problem that we can't guarantee than analytics is alive if main thread alive.
This should fix https://github.com/hastic/hastic-server/issues/316 and https://github.com/hastic/hastic-server/issues/407 in theory, but load test needed to check

## Implementation 

`ServerService` now is `AsyncZmqActor` so it works is a separate thread.

### DIFF
--- a/analytics/analytics/services/server_service.py
+++ b/analytics/analytics/services/server_service.py
@@ -7,6 +7,9 @@ import logging
 import json
 import asyncio
 import traceback
+
+import utils.concurrent
+
 from typing import Optional
 
 logger = logging.getLogger('SERVER_SERVICE')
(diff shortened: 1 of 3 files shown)


analytics/analytics/services/server_service.py
```diff
@@ -39,60 +42,75 @@ def fromJSON(json: dict):
...
+    async def _on_message_to_thread(self, message: str):
+        await self.__server_socket.send_string(message)
+
+    async def __server_socket_recv_loop(self):
+        while True:
+            received_string = await self.__server_socket.recv_string()
+            if received_string == 'PING':
+                asyncio.ensure_future(self.__handle_ping())
+            else:
+                self._send_message_from_thread(received_string)
```
rozetko: Why don't we wrap it in `asyncio.ensure_future`?

analytics/analytics/services/server_service.py
```diff
@@ -39,60 +42,75 @@ def fromJSON(json: dict):
...
 
     async def __handle_ping(self):
-        await self.socket.send(b'PONG')
+        await self.__server_socket.send(b'PONG')
```
rozetko: Maybe we can do `send_string('PONG')` here?

analytics/analytics/services/server_service.py
```diff
@@ -39,61 +46,79 @@ def fromJSON(json: dict):
...
+        # self.__actor_socket.send_string(json.dumps(message.toJSON()))
+        await self._put_message_to_thread(json.dumps(message.toJSON()))
+
+    async def send_request_to_server(self, message: ServerMessage) -> object:
+        if message.request_id is not None:
+            raise ValueError('Message can`t have request_id before it is scheduled')
+        request_id = message.request_id = self.__request_next_id
+        self.request_next_id = self.__request_next_id + 1
+        asyncio.ensure_future(self.send_message_to_server(message))
+        while request_id not in self.__responses:
```
jonyrock: you should `await self.__responses[request_id]` which should be a task, which you resolve somewhere else

jonyrock:
```suggestion
        # you should await self.__responses[request_id] which should be a task, 
        # which you resolve somewhere else
        while request_id not in self.__responses:
```

PR #532, opened 2019-04-08
Dataframe for detection less than two window size

### LINKED ISSUE #531: Dataframe for detection less than two window size
Check that dataframe length is at least 2 window sizes

amper43:
fixes #531 

Changes:
- added check that data length at least two window size
- added test


analytics/analytics/detectors/pattern_detector.py
```diff
@@ -57,6 +57,17 @@ def detect(self, dataframe: pd.DataFrame, cache: Optional[models.ModelCache]) ->
         logger.debug('Unit {} got {} data points for detection'.format(self.analytic_unit_id, len(dataframe)))
         # TODO: split and sleep (https://github.com/hastic/hastic-server/pull/124#discussion_r214085643)
 
+        window_size = cache.get('WINDOW_SIZE')
```
rozetko: Let's add a test for `cache = {}`

amper43: @rozetko  worker handles this case

jonyrock:
data length where?

Issue #518, opened 2019-04-05
ValueError: attempt to get argmin of an empty sequence


rozetko:
Error while learning:
```
analytics_1  |   File "/usr/local/lib/python3.6/asyncio/tasks.py", line 358, in wait_for
analytics_1  |     return fut.result()
analytics_1  |   File "/usr/local/lib/python3.6/concurrent/futures/thread.py", line 56, in run
analytics_1  |     result = self.fn(*self.args, **self.kwargs)
analytics_1  |   File "bin/../analytics/detectors/pattern_detector.py", line 49, in train
analytics_1  |     new_cache = self.model.fit(dataframe, segments, self.analytic_unit_id, cache)
analytics_1  |   File "bin/../analytics/models/model.py", line 88, in fit
analytics_1  |     learning_info = self.get_parameters_from_segments(dataframe, labeled, deleted, model, model_type)
analytics_1  |   File "bin/../analytics/models/model.py", line 149, in get_parameters_from_segments
analytics_1  |     learning_info['pattern_width'].append(utils.find_width(aligned_segment, model_type))
analytics_1  |   File "bin/../analytics/utils/common.py", line 176, in find_width
analytics_1  |     left_extremum_index = utils.find_last_extremum(pattern_left, selector)
analytics_1  |   File "bin/../analytics/utils/common.py", line 184, in find_last_extremum
analytics_1  |     first_extremum_ind = find_extremum_index(segment, not selector)
analytics_1  |   File "bin/../analytics/utils/common.py", line 192, in find_extremum_index
analytics_1  |     return segment.argmin()
analytics_1  | ValueError: attempt to get argmin of an empty sequence
```

PR #537, opened 2019-04-08
ValueError attempt to get argmin of an empty sequence #518

VargBurz:
#518


analytics/analytics/models/model.py
```diff
@@ -143,10 +143,13 @@ def get_parameters_from_segments(self, dataframe: pd.DataFrame, labeled: list, d
             learning_info['pattern_timestamp'].append(segment.pattern_timestamp)
             aligned_segment = utils.get_interval(data, segment_center, self.state['WINDOW_SIZE'])
             aligned_segment = utils.subtract_min_without_nan(aligned_segment)
+            if len(aligned_segment) == 0:
+                logging.warning('cant add segment to learning because segment is empty with center: {}, window_size: {}, and len_data: {}'.format(
```
jonyrock: veeery log line

jonyrock: ".... segment is empty with center" -> "... segment is empty where segment's center is .... "

PR #539, opened 2019-04-08
VSCode remote debugging via ptvsd

amper43:
For remote debugging:
- make ssh tunell via `ssh -fNL 5679:localhost:5678 <server>` (you should have ssh config with `ServerAliveInterval`)
- run in vscode debug tab `attach to remote` for analytics

Changes:
- added configuration for remote debugging from vscode
- added pip package `ptvsd` for vscode debugging

### DIFF
--- a/analytics/requirements.txt
+++ b/analytics/requirements.txt
@@ -16,3 +16,4 @@ scipy==1.1.0
 seglearn==0.1.6
 six==1.11.0
 tsfresh==0.11.0
+ptvsd==4.2.6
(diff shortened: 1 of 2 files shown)


analytics/requirements.txt
```diff
@@ -16,3 +16,4 @@ scipy==1.1.0
 seglearn==0.1.6
 six==1.11.0
 tsfresh==0.11.0
+ptvsd==4.2.6
```
rozetko: Maybe it could be in `dev_requierements.txt`, for example?

PR #558, opened 2019-04-11
Hastic instance name

### LINKED ISSUE #547: Webhooks: Hastic instance name / url
Now you don't know which Hastic instance is sending you a webhook
Though it can be critical in cases like `Analytics is down`
We should add a field with some hastic-server identifier

amper43:
fixes #547 

changes:
- added `instanceName` field to server info and webhook payload
- added label `instanceName` to `prometheus-hastic-exporter`

### DIFF
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -35,6 +35,7 @@ export const INSIDE_DOCKER = process.env.INSIDE_DOCKER !== undefined;
 export const PRODUCTION_MODE = process.env.NODE_ENV !== 'development';
 
 export const ZMQ_CONNECTION_STRING = createZMQConnectionString();
+export const HASTIC_INSTANCE_NAME = getConfigField('HASTIC_INSTANCE_NAME', null);
 
 
 function getConfigField(field: string, defaultVal?: any) {
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -35,7 +35,7 @@ export const INSIDE_DOCKER = process.env.INSIDE_DOCKER !== undefined;
 export const PRODUCTION_MODE = process.env.NODE_ENV !== 'development';
 
 export const ZMQ_CONNECTION_STRING = createZMQConnectionString();
-export const HASTIC_INSTANCE_NAME = getConfigField('HASTIC_INSTANCE_NAME', null);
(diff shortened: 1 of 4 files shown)


server/src/services/notification_service.ts
```diff
@@ -1,9 +1,10 @@
 import { Segment } from '../models/segment_model';
 import * as AnalyticUnit from '../models/analytic_unit_model';
-import { HASTIC_WEBHOOK_URL, HASTIC_WEBHOOK_TYPE, HASTIC_WEBHOOK_SECRET } from '../config';
+import { HASTIC_WEBHOOK_URL, HASTIC_WEBHOOK_TYPE, HASTIC_WEBHOOK_SECRET, HASTIC_INSTANCE_NAME } from '../config';
 
 import axios from 'axios';
 import * as querystring from 'querystring';
+import * as _ from 'lodash';
```
rozetko: Why do we need `lodash` here?

server/src/config.ts
```diff
@@ -35,6 +35,7 @@ export const INSIDE_DOCKER = process.env.INSIDE_DOCKER !== undefined;
 export const PRODUCTION_MODE = process.env.NODE_ENV !== 'development';
 
 export const ZMQ_CONNECTION_STRING = createZMQConnectionString();
+export const HASTIC_INSTANCE_NAME = getConfigField('HASTIC_INSTANCE_NAME', null);
```
rozetko: Maybe we should make default value `'unset'`, for example? Instead of null

rozetko: Bu I don't know if it's a good idea
What do you think?

jonyrock: or
```
var os = require("os");
os.hostname();
```

amper43: @jonyrock hostname often returns `localhost`

server/src/services/notification_service.ts
amper43:
```suggestion
```

server/src/config.ts
jonyrock: @amper43 still better than `undefined`

rozetko: +1 for `os.hostname()`

PR #564, opened 2019-04-12
Endpoint for getting timeseries of analytic unit #528

### LINKED ISSUE #528: Endpoint for getting timeseries of analytic unit
Imagine that you want to export data from datasource by Hastic. 
... or you want to see in HasticApp timeseries, but from Hastic, not from Grafana
... or you want to debug dataflow and see how exactly node-server returns data to analytics

amper43:
fixes #528 

changes:
- added `data_router.ts` with `query` that just proxy for data-kit

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -1,6 +1,7 @@
 import { router as analyticUnitsRouter } from './routes/analytic_units_router';
 import { router as segmentsRouter } from './routes/segments_router';
 import { router as thresholdRouter } from './routes/threshold_router';
+import { router as dataRouter } from './routes/data_router';
 
 import * as AnalyticsController from './controllers/analytics_controller';
 
@@ -55,6 +56,7 @@ async function init() {
   rootRouter.use('/analyticUnits', analyticUnitsRouter.routes(), analyticUnitsRouter.allowedMethods());
   rootRouter.use('/segments', segmentsRouter.routes(), segmentsRouter.allowedMethods());
   rootRouter.use('/threshold', thresholdRouter.routes(), thresholdRouter.allowedMethods());
+  rootRouter.use('/data', dataRouter.routes(), dataRouter.allowedMethods());
 
   rootRouter.get('/', async (ctx) => {
     const activeWebhooks = await AnalyticsController.getActiveWebhooks();
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -56,7 +56,7 @@ async function init() {
   rootRouter.use('/analyticUnits', analyticUnitsRouter.routes(), analyticUnitsRouter.allowedMethods());
   rootRouter.use('/segments', segmentsRouter.routes(), segmentsRouter.allowedMethods());
   rootRouter.use('/threshold', thresholdRouter.routes(), thresholdRouter.allowedMethods());
-  rootRouter.use('/data', dataRouter.routes(), dataRouter.allowedMethods());
+  rootRouter.use('/query', dataRouter.routes(), dataRouter.allowedMethods());
 
   rootRouter.get('/', async (ctx) => {
     const activeWebhooks = await AnalyticsController.getActiveWebhooks();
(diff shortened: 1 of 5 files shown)


server/src/routes/data_router.ts
```diff
@@ -0,0 +1,61 @@
...
+    const message = `analytic unit with id ${analyticUnitId} not present in data base`;
+    console.error(message);
+    throw new Error(message);
+  }
+
+  let grafanaUrl;
+  if(GRAFANA_URL !== null) {
+    grafanaUrl = GRAFANA_URL;
+  } else {
+    grafanaUrl = analyticUnit.grafanaUrl;
```
jonyrock: I don't know for sure, but does `analyticUnit.grafanaUrl` have less priority than `GRAFANA_URL` ?

server/src/index.ts
```diff
@@ -55,6 +56,7 @@ async function init() {
   rootRouter.use('/analyticUnits', analyticUnitsRouter.routes(), analyticUnitsRouter.allowedMethods());
   rootRouter.use('/segments', segmentsRouter.routes(), segmentsRouter.allowedMethods());
   rootRouter.use('/threshold', thresholdRouter.routes(), thresholdRouter.allowedMethods());
+  rootRouter.use('/data', dataRouter.routes(), dataRouter.allowedMethods());
```
jonyrock: let's call this endpoint similar to Grafana endpoint which queries data?

server/src/routes/data_router.ts
```diff
@@ -0,0 +1,61 @@
+import * as AnalyticUnit from '../models/analytic_unit_model';
+import { queryByMetric, GrafanaUnavailable, DatasourceUnavailable } from 'grafana-datasource-kit';
+import { HASTIC_API_KEY, GRAFANA_URL } from '../config';
```
jonyrock: I think it should be above `'grafana-datasource-kit'` cuz it's more specific

server/src/routes/data_router.ts
```diff
@@ -0,0 +1,61 @@
...
+  let grafanaUrl;
+  if(GRAFANA_URL !== null) {
+    grafanaUrl = GRAFANA_URL;
+  } else {
+    grafanaUrl = analyticUnit.grafanaUrl;
+  }
+
+  try {
+    const data = await queryByMetric(analyticUnit.metric, grafanaUrl, from, to, HASTIC_API_KEY);
+    ctx.response.body = { data };
```
jonyrock: let's make it similar to Grafana response

server/src/routes/data_router.ts
rozetko: Yes, global `GRAFANA_URL` has higher priority
That's because we get `analyticUnit.grafanaUrl` from browser address bar and it can be unreachable from Hastic

rozetko: BTW, I've seen this code already
Is it duplicated? Maybe it would be better to move it to some function?

server/src/routes/data_router.ts
```diff
@@ -0,0 +1,61 @@
...
+  } catch(e) {
+    console.error(`data router got exception ${e} while query data`);
```
rozetko: Don't know if it's a good idea to log error because it's already logged in middleware
https://github.com/hastic/hastic-server/blob/master/server/src/index.ts#L43-L51

server/src/routes/data_router.ts
```diff
@@ -0,0 +1,61 @@
...
+    const message = `data router error: 'to' must be greater than 'from' (from:${from} to:${to})`;
+    console.error(message);
+    throw new Error(message);
+  }
+
+  const analyticUnit = await AnalyticUnit.findById(analyticUnitId);
+
+  if(analyticUnit === undefined) {
+    const message = `analytic unit with id ${analyticUnitId} not present in data base`;
+    console.error(message);
```
rozetko: Not neccessary to log it here
We can just `throw`

server/src/routes/data_router.ts
```diff
@@ -0,0 +1,61 @@
...
+  if(to === undefined) {
+    throw new Error(`data router error: request must contain 'to'`)
+  }
+
+  from = +from;
+  to = +to;
+
+  if(to <= from) {
+    const message = `data router error: 'to' must be greater than 'from' (from:${from} to:${to})`;
+    console.error(message);
```
rozetko: Same here

server/src/routes/data_router.ts
amper43: @jonyrock in Grfana's responce metric's data presents in `data` field

PR #573, opened 2019-04-15
Analytic types

jonyrock:
Let's make one place where we define common for hastic analytics types. 
Also, we have two definitions of `AnalyticUnitId = str`


jonyrock:
@amper43 I added a comment to `/__init__.py`

PR #583, opened 2019-04-16
jsonclass usage in models + fixes in meta

jonyrock:
We have a nice decorator `utils.meta.JSONClass` which we can use in models

also I fixed small bugs like static field usage and `super()` incorrect call

### DIFF
--- a/analytics/analytics/models/general_model.py
+++ b/analytics/analytics/models/general_model.py
@@ -1,40 +1,34 @@
+from analytic_types import AnalyticUnitId
 from models import Model, ModelState
 from typing import Union, List, Generator
 import utils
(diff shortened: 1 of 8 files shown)


analytics/analytics/models/general_model.py
```diff
@@ -1,40 +1,34 @@
...
 from scipy.stats.stats import pearsonr
-from typing import Optional
-import math
+
 from scipy.stats import gaussian_kde
 from scipy.stats import norm
 import logging
-from analytic_types import AnalyticUnitId
 
+from typing import Optional
```
amper43: unused imports

PR #589, opened 2019-04-19
Add proxy class for getting segments #588

### LINKED ISSUE #588: Add proxy class for getting segments
Subtask for #504 

Firstly, we should add a class which proxies queries for getting segments to detection

amper43:
fixes #588 

changes:
- added detection_status_router that proxy calls to `runDetect` of `analytic_controller`
- `runDetect` signature now can recieve `from` & `to` to set time range for detection
- `detection_status_router`  now listen `/detectionStatus`


server/src/controllers/analytics_controller.ts
```diff
@@ -245,7 +246,11 @@ export async function runDetect(id: AnalyticUnit.AnalyticUnitId) {
     let analyticUnitType = unit.type;
     let detector = AnalyticUnit.getDetectorByType(analyticUnitType);
 
-    const data = await query(unit, detector);
+    let range = undefined;
```
```suggestion
    let range;
```

server/src/routes/detection_status_router.ts
```diff
@@ -0,0 +1,65 @@
...
+import * as Router from 'koa-router';
+import * as _ from 'lodash';
+
+export enum DetectionStatus {
+  READY = 'READY',
+  RUNNING = 'RUNNING',
+  FAILED = 'FAILED'
+}
+
+declare type RunnedDetection = {
```
rozetko: this and `DetectionStatus` would be better moved to models

Issue #592, opened 2019-04-20
Check python version in run time


jonyrock:
User may try to start analytics with python version less that required 3.6.5
we need to check it.

jonyrock:
something like https://hg.python.org/cpython/file/23a60d89dbd4/Lib/unittest.py#l70

jonyrock:
https://github.com/hastic/hastic-server/pull/595

PR #595, opened 2019-04-22
Check python version in run time #592

jonyrock:
I had to move most of the logic from `bin/server` to a new file `analytics/server`

Then I added python check

### DIFF
--- a/analytics/bin/server
+++ b/analytics/bin/server
@@ -3,107 +3,26 @@
 import sys
 import os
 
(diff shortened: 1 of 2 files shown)


analytics/bin/server
```diff
@@ -3,107 +3,30 @@
 import sys
 import os
 
-#TODO: make wrapper script that set PYTHONPATH instead
+if sys.version_info[:3] < (3, 6, 5):
```
rozetko: Also, version can't be >= 3.7

PR #611, opened 2019-04-29
Exponential smoothing for anomaly detector #610

### LINKED ISSUE #610: Exponential smoothing for anomaly detector
Need to refresh exponential smoothing function in utils so anomaly detector could use it

VargBurz:
closes #610


analytics/analytics/utils/common.py
```diff
@@ -16,15 +16,23 @@
...
 
-def exponential_smoothing(series, alpha):
-    result = [series[0]]
+
+def exponential_smoothing(series: pd.Series, alpha: float, last_smooth: Optional[float] = None) -> pd.Series:
+    if alpha < 0 or alpha > 1:
+        raise ValueError('Alpha must be within the boundaries: 0 <= alpha <= 1')
+    if len(series) < 2:
+        return series
+    if not last_smooth:
```
```suggestion
    if last_smooth == None:
```

analytics/analytics/utils/common.py
```diff
@@ -16,15 +16,23 @@
 CONFIDENCE_FACTOR = 0.5
 SMOOTHING_FACTOR = 5
 
-def exponential_smoothing(series, alpha):
-    result = [series[0]]
+
+def exponential_smoothing(series: pd.Series, alpha: float, last_smooth: Optional[float] = None) -> pd.Series:
```
rozetko: `last_smooth` -> `last_smoothed_value`

analytics/analytics/utils/common.py
```diff
@@ -16,15 +16,23 @@
...
+
+def exponential_smoothing(series: pd.Series, alpha: float, last_smoothed_value: Optional[float] = None) -> pd.Series:
+    if alpha < 0 or alpha > 1:
+        raise ValueError('Alpha must be within the boundaries: 0 <= alpha <= 1')
+    if len(series) < 2:
+        return series
+    if last_smoothed_value is None:
+        result = [series.values[0]]
+    else:
+        result = [float(last_smoothed_value)]
```
rozetko: Do we need `last_smoothed_value` in result?
It isn't part of `series`

VargBurz: if 'last_smoothed_value' get in func we should use it.
If not  - can use the standard method

rozetko: I mean, we can use it to generate new values but don't include it in result

PR #620, opened 2019-05-01
Threshold is not a part of analytic unit #619

### LINKED ISSUE #619: Threshold is not a part of analytic unit
Threshold config is stored and handled separately from analytic unit 
We should make children `AnalyticUnit` classes for different types of detectors (e.g. `PatternAnalyticUnit`, `ThresholdAnalyticUnit`)
It would make anomaly detector adding (#606) easier

rozetko:
Fixes #619 

Changes:
- Add children `AnalyticUnit` classes for each detector type
- Integrate `thresholds` into analytic unit (instead of using it as a separate entity)
- Change `models/analytic_units` directory structure (`index.ts` exports everything from other analytic_unit modules) :
image
- Add migration


server/src/controllers/analytics_controller.ts
```diff
@@ -226,8 +226,10 @@ export async function runLearning(id: AnalyticUnit.AnalyticUnitId) {
       segmentObjs = _.concat(segmentObjs, deletedSegmentsObjs);
       taskPayload.segments = segmentObjs;
     } else if(detector === AnalyticUnit.DetectorType.THRESHOLD) {
-      const threshold = await Threshold.findOne(id);
-      taskPayload.threshold = threshold;
+      taskPayload.threshold = {
+        value: (<ThresholdAnalyticUnit>analyticUnit).value,
```
jonyrock: it's always better to use `as` keyword for casting

server/src/models/analytic_units/analytic_unit_model.ts
```diff
@@ -0,0 +1,75 @@
...
+import { Metric } from 'grafana-datasource-kit';
+
+
+export abstract class AnalyticUnit {
+  constructor(
+    public name: string,
+    public grafanaUrl: string,
+    public panelId: string,
+    // TODO: enum type
+    // TODO: type -> modelType
```
rozetko:
```suggestion
    // TODO: type -> subType
```

PR #622, opened 2019-05-02
Add migrations how-to

rozetko:
Added comment about adding new migration


server/src/migrations.ts
```diff
@@ -1,3 +1,10 @@
+/*
+  How to add migrations:
```
jonyrock:
```suggestion
  How to add a migration:
```

PR #623, opened 2019-05-02
Thresholds error: from is NaN #621

### LINKED ISSUE #621: Thresholds error: from is NaN
You get this error any time you click `Apply` on threshold analytic unit
image
There was no error before implementing https://github.com/hastic/hastic-server/issues/504

rozetko:
Fixes #621 

The error occurred because of missing `windowSize` field in Threshold`s detection cache 

Changes:
- use `from` and `to` in learning and detection if they're provided 
- check if `cache` has `windowSize`
- remove previously found segments before learning

### DIFF
--- a/server/src/controllers/analytics_controller.ts
+++ b/server/src/controllers/analytics_controller.ts
@@ -474,6 +474,10 @@ export async function runLearningWithDetection(
 ): Promise<void> {
   // TODO: move setting status somehow "inside" learning
   await AnalyticUnit.setStatus(id, AnalyticUnit.AnalyticUnitStatus.PENDING);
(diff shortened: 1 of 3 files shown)


server/src/controllers/analytics_controller.ts
```diff
@@ -232,7 +232,12 @@ export async function runLearning(id: AnalyticUnit.AnalyticUnitId) {
       };
     }
 
-    const range = await getQueryRange(id, detector);
+    let range;
```
jonyrock: can we define TimeRange type somewhere in the code, please?

jonyrock: it's cuz u call `query(analyticUnit, range)` and technically, u don't define `range` type for this variable.

PR #641, opened 2019-05-09
HASTIC_PORT in .env file #640

### LINKED ISSUE #640: HASTIC_PORT in .env file
If you run hastic-server using docker-compose, the only way to change port is editing `docker-compose.yml`. It may be inconvenient because `docker-compose.yml` is under git
We should make it possible to set hastic-server port in `.env` file

rozetko:
Closes #640 

Now it's possible to set `HASTIC_PORT` variable in `.env` file
Compatible with old configurations: default value is 8000


jonyrock:
@amper43 please just take a look

PR #645, opened 2019-05-12
VSCode settings improvements in analytics

jonyrock:
[VSCode](https://code.visualstudio.com) is the editor we use the most and it's a big part of our workflow. 
This PR aggregates many small changes made in `analytics/.vscode/launch.json` and `analytics/.vscode/settings.json` files.

## Changes
* Format `launch.json` and `settings.json` with 2 spaces instead of 4:
we format ALL `*.json` (as well as `*.ts` / `.js`) files with 2 spaces
* In `settings.json`, set `"editor.tabSize": 4` only for `*.py` (python) files:
we need such formatting only there
* Exclude folders `dist` and `build` from project
* Rename `python.unitTest.*` to `python.testing.*` bacause of error: 

* Usage of `PYTHONPATH` in `.vscode/.env` (see following paragraph)

### PYTHONPATH

In `bin/server` we have following code: 
```python
sys.path.append(os.path.join(os.path.dirname(__file__), '..', 'analytics'))
```

but VSCode can't resolve this and shows import error:

To resolve this, environment variable `PYTHONPATH` now defined in `.vscode/.env`. I put `.env` file into `.vscode` because in the root folder it looks suspicious: only vscode uses it.

From docs: https://code.visualstudio.com/docs/python/environments#_use-of-the-pythonpath-variable important note:

> **Note**: PYTHONPATH does **not**, repeat **not**, specify a path to a Python interpreter itself, and thus you **never** use it with the `python.pythonPath` setting. Clearly, the environment variable was badly named, but...[c'est la vie](https://www.dictionary.com/browse/c-est-la-vie). So make sure to read the [PYTHONPATH documentation](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH) several times and fix in your mind that PYTHONPATH is **not** a path to an interpreter.

## See also
* https://code.visualstudio.com/docs/getstarted/settings
* https://code.visualstudio.com/docs/python/settings-reference
* https://code.visualstudio.com/docs/python/unit-testing
* https://code.visualstudio.com/docs/python/environments#_use-of-the-pythonpath-variable
* https://github.com/Microsoft/vscode-python/issues/3840#issuecomment-463789294
* https://code.visualstudio.com/docs/python/environments#_choose-a-debugging-environment

### DIFF
--- a/analytics/.vscode/.env
+++ b/analytics/.vscode/.env
@@ -1,0 +1,1 @@
+PYTHONPATH=analytics
--- a/analytics/.vscode/settings.json
+++ b/analytics/.vscode/settings.json
@@ -1,18 +1,23 @@
-{
-  "python.pythonPath": "python",
-  "python.linting.enabled": true,
+{  
   "terminal.integrated.shell.windows": "C:\\WINDOWS\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",
-  "editor.tabSize": 4,
   "editor.insertSpaces": true,
   "files.eol": "\n",
   "files.exclude": {
     "**/__pycache__/": true
-},
-"python.unitTest.unittestArgs": [
+  },
+
+  "[python]": {
+    "editor.tabSize": 4,
+  },
+
+  "python.pythonPath": "python",
+  "python.linting.enabled": true,
+
+  "python.unitTest.unittestArgs": [
     "-v"
-],
-"python.unitTest.pyTestEnabled": false,
-"python.unitTest.nosetestsEnabled": false,
-"python.unitTest.unittestEnabled": true,
-"python.linting.pylintEnabled": true
+  ],
+  "python.unitTest.pyTestEnabled": false,
+  "python.unitTest.nosetestsEnabled": false,
+  "python.unitTest.unittestEnabled": true,
+  "python.linting.pylintEnabled": true
 }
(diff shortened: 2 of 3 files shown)


analytics/.vscode/.env
```diff
@@ -0,0 +1 @@
+PYTHONPATH=analytics
```
rozetko: Is this path relative to workspace root?

analytics/.vscode/settings.json
```diff
@@ -1,18 +1,21 @@
...
-"python.linting.pylintEnabled": true
+    "**/__pycache__/": true,
+    "dist": true,
+    "build": true
+  },
+  "[python]": {
+    "editor.tabSize": 4,
+  },
+  "python.envFile": "${workspaceFolder}/.vscode/.env",
+  "python.pythonPath": "python",
```
rozetko: Maybe it'd be better to make it different for windows and linux?
For example:
```json
"windows": {
  "python.pythonPath": "python"
},
"linux": {
  "python.pythonPath": "python3"
}
```

amper43: yes, could you please set `python3` for linux because `python` on ubuntu 16  refers to `python2.7`

analytics/.vscode/.env
jonyrock: Yes, it is, from projet root

jonyrock:
We wanted also to add 

* Windows/Linux `python.pythonPath` setting: `python` for windows and `python3` for Linux

but I haven't found a solution for this.

PR #646, opened 2019-05-13
Merge threshold segments #624

### LINKED ISSUE #624: Merge threshold segments
Each threshold segment contains only one point
We should merge them

VargBurz:
Closes #624
## Problem
After the detection of threshold neighboring segments that were found did not merge. The reason - timestamps differ from each other at some step.
image

## Changes
- Add method `get_time_step` that defines the time step(the difference between neighbor values' timestamps)
- Method `merge_intersecting_segments` uses `time_step`
- Worker  call method  `get_time_step` before merging segments

image

### DIFF
--- a/analytics/analytics/analytic_types/__init__.py
+++ b/analytics/analytics/analytic_types/__init__.py
@@ -17,7 +17,7 @@ AnalyticUnitId = str
 ModelCache = dict
 
 # TODO: explicit timestamp / value
-TimeSeries = [List[Tuple[int, int]]]
+TimeSeries = [List[Tuple[int, float]]]
 
 """
 Example:
--- a/analytics/analytics/analytic_types/__init__.py
+++ b/analytics/analytics/analytic_types/__init__.py
@@ -17,7 +17,7 @@ AnalyticUnitId = str
 ModelCache = dict
 
 # TODO: explicit timestamp / value
-TimeSeries = [List[Tuple[int, float]]]
+TimeSeries = List[Tuple[int, float]]
 
 """
 Example:
--- a/analytics/analytics/utils/common.py
+++ b/analytics/analytics/utils/common.py
@@ -128,7 +128,7 @@ def close_filtering(pattern_list: List[int], win_size: int) -> List[Tuple[int, i
             s.append([pattern_list[i]])
     return s
 
-def merge_intersecting_segments(segments: List[Segment]) -> List[Segment]:
+def merge_intersecting_segments(segments: List[Segment], time_step: int) -> List[Segment]:
     '''
     Find intersecting segments in segments list and merge it.
     '''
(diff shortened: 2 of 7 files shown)


analytics/analytics/utils/common.py
```diff
@@ -137,14 +137,19 @@ def merge_intersecting_segments(segments: List[Segment]) -> List[Segment]:
...
-        if segments[i].from_timestamp <= previous_segment.to_timestamp:
+        if segments[i].from_timestamp <= previous_segment.to_timestamp + time_step:
             segments[i].from_timestamp = min(previous_segment.from_timestamp, segments[i].from_timestamp)
             segments[i].to_timestamp = max(previous_segment.to_timestamp, segments[i].to_timestamp)
             segments[i - 1] = None
         previous_segment = segments[i]
     segments = [x for x in segments if x is not None]
     return segments
 
+def get_time_step(dataframe: pd.DataFrame) -> int:
```
jonyrock:
```suggestion
def find_interval(dataframe: pd.DataFrame) -> int:
```

analytics/analytics/utils/common.py
```diff
@@ -137,14 +137,19 @@ def merge_intersecting_segments(segments: List[Segment]) -> List[Segment]:
...
+    if len(dataframe) < 2:
+        raise ValueError('Couldnt get time step from dataframe: {}'.format(dataframe))
```
jonyrock: `'Couldnt find interval from dataframe:`

analytics/analytics/utils/common.py
```diff
@@ -137,14 +137,19 @@ def merge_intersecting_segments(segments: List[Segment]) -> List[Segment]:
...
             segments[i].to_timestamp = max(previous_segment.to_timestamp, segments[i].to_timestamp)
             segments[i - 1] = None
         previous_segment = segments[i]
     segments = [x for x in segments if x is not None]
     return segments
 
+def find_interval(dataframe: pd.DataFrame) -> int:
+    if len(dataframe) < 2:
+        raise ValueError('Can`t find interval: length of dataframe must be at least 2')
+    return int(dataframe[1][0] - dataframe[0][0])
```
rozetko: Maybe like this?
```suggestion
    return int(dataframe['timestamp'][0] - dataframe['timestamp'][1])
```

VargBurz: No, it is not dataframe. I changed type to `List[Tuple[int, float]]`

rozetko: Let's use `TimeSeries` from `analytic_typing` instead of `List[Tuple[int, float]]`
It's in master

analytics/analytics/analytic_types/__init__.py
```diff
@@ -17,7 +17,7 @@
 ModelCache = dict
 
 # TODO: explicit timestamp / value
-TimeSeries = [List[Tuple[int, int]]]
+TimeSeries = [List[Tuple[int, float]]]
```
rozetko:
```suggestion
TimeSeries = List[Tuple[int, float]]
```

analytics/analytics/analytic_types/__init__.py
```diff
@@ -17,7 +17,7 @@
 ModelCache = dict
 
 # TODO: explicit timestamp / value
-TimeSeries = [List[Tuple[int, int]]]
+TimeSeries = List[Tuple[int, float]]
```
jonyrock: the only problem is that you already have `TimeSeries` class in [line 34](https://github.com/hastic/hastic-server/pull/646/files#diff-c908fa5ba59b7215e8324a45f1f9e440R34)

jonyrock: but I will rename the class below by myself soon, but be accurate plz

PR #668, opened 2019-05-20
AnomalyDetector: Need at least 1 labeled segment #667

### LINKED ISSUE #667: AnomalyDetector: Need at least 1 labeled segment
When you click `Re-detect all analytic units` in panel, you get "Need at least 1 labeled segment" error for anomaly analytic units without seasonality. Although they don't need segments for learning

VargBurz:
Closes #667 

## Problem:

When you click `re-detect all analytic units` in panel, you get "Need at least 1 labeled segment" error for anomaly analytic units without seasonality. Although they don't need segments for learning.

## Change:

Add separate condition for anomaly detector in `getQueryRange`. 
If analytic unit has segments - method uses pattern detector logic,
if there are no segments - method uses threshold detector logic.

### DIFF
--- a/server/src/controllers/analytics_controller.ts
+++ b/server/src/controllers/analytics_controller.ts
@@ -152,6 +152,7 @@ async function getQueryRange(
     default:
       throw new Error(`Cannot get query range for detector type ${detectorType}`);
   }
+}
 
 async function query(
   analyticUnit: AnalyticUnit.AnalyticUnit,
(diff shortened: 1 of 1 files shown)


server/src/controllers/analytics_controller.ts
```diff
@@ -140,6 +139,20 @@ async function getQueryRange(
     };
   }
 
+  if (detectorType === AnalyticUnit.DetectorType.ANOMALY) {
+    const segments = await Segment.findMany(analyticUnitId, { $or: { labeled: true, deleted: true } });
+    if (segments.length === 0) {
```
```suggestion
    if(segments.length === 0) {
```

server/src/controllers/analytics_controller.ts
```diff
@@ -140,6 +139,20 @@ async function getQueryRange(
     };
   }
 
+  if (detectorType === AnalyticUnit.DetectorType.ANOMALY) {
```
rozetko: Let's use `switch(detectorType)` instead of `if`s

PR #670, opened 2019-05-20
Anomaly detector webhooks fix

amper43:
## Changes

- calculate time step for merging results in detector's  `detect()` method
- fix utils method `find_interval`

### DIFF
--- a/analytics/analytics/utils/common.py
+++ b/analytics/analytics/utils/common.py
@@ -149,7 +149,7 @@ def merge_intersecting_segments(segments: List[Segment], time_step: int) -> List
 def find_interval(dataframe: pd.DataFrame) -> int:
     if len(dataframe) < 2:
         raise ValueError('Can`t find interval: length of data must be at least 2')
-    delta = int(dataframe.timestamp[1]) - int(dataframe.timestamp[0])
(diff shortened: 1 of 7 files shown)


analytics/analytics/utils/common.py
```diff
@@ -146,10 +146,11 @@ def merge_intersecting_segments(segments: List[Segment], time_step: int) -> List
     segments = [x for x in segments if x is not None]
     return segments
 
-def find_interval(data: TimeSeries) -> int:
-    if len(data) < 2:
+def find_interval(dataframe: pd.DataFrame) -> int:
+    if len(dataframe) < 2:
         raise ValueError('Can`t find interval: length of data must be at least 2')
-    return int(data[1][0] - data[0][0])
+    delta = int((dataframe.timestamp[1].timestamp() - dataframe.timestamp[0].timestamp()) * 1000)
```
rozetko: Maybe it would be better to use `utils.convert_pd_timestamp_to_ms(dataframe['timestamp'][0])` instead of `dataframe.timestamp[0].timestamp()`)
1. We do it everywhere
2. We wouldn't need to multiply it by 1000

rozetko: We can just add a TODO now

PR #682, opened 2019-05-27
Change mutable default values to none #638

### LINKED ISSUE #638: Change mutable default values to None
We should do:
```python
def myfunc(value = None):
    if value is None:
        value = []
    # modify value here
```
Instead of:
```python
def myfunc(value = []):
    # modify value here
```
More info here: http://effbot.org/zone/default-values.htm

amper43:
fixes #638 

### Changes
- removed mutables from defaults in `ModelState`
- added types
- removed unused method `segments_box`

### DIFF
--- a/analytics/analytics/analytic_unit_manager.py
+++ b/analytics/analytics/analytic_unit_manager.py
@@ -44,7 +44,7 @@ class AnalyticUnitManager:
         self.analytic_workers[analytic_unit_id] = worker
         return worker
 
-    async def __handle_analytic_task(self, task) -> dict:
+    async def __handle_analytic_task(self, task: object) -> dict:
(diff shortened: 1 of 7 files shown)


analytics/analytics/analytic_unit_manager.py
```diff
@@ -44,7 +44,7 @@ def __ensure_worker(
         self.analytic_workers[analytic_unit_id] = worker
         return worker
 
-    async def __handle_analytic_task(self, task) -> dict:
+    async def __handle_analytic_task(self, task: object) -> dict:
```
jonyrock: I have a dream that one day we will stop using `object` and `dict` types in our or code.

PR #702, opened 2019-06-19
WIP: Detection re-runs in RUNNING DetectionStatus #699

### LINKED ISSUE #699: Detection re-runs in RUNNING DetectionStatus
If timerange is in `RUNNING` DetectionStatus, detection re-runs on `/span` query. Thus, you can get infinite `RUNNING` status.
It should just return `RUNNING` instead.
P.S. maybe the reason is span extending?

amper43:
fixes #699 

## Changes
- filtering spans that inside or intersected with `FAILED` spans

### DIFF
--- a/server/src/controllers/analytics_controller.ts
+++ b/server/src/controllers/analytics_controller.ts
@@ -18,6 +18,7 @@ import { queryByMetric, GrafanaUnavailable, DatasourceUnavailable } from 'grafan
 import * as _ from 'lodash';
 import { WebhookType } from '../services/notification_service';
 import { AnomalyAnalyticUnit } from '../models/analytic_units/anomaly_analytic_unit_model';
+import { read } from 'fs';
 
 const SECONDS_IN_MINUTE = 60;
 
@@ -399,7 +400,9 @@ export async function runDetect(id: AnalyticUnit.AnalyticUnitId, from?: number, 
         id,
         range.from + intersection,
         range.to - intersection,
-        Detection.DetectionStatus.FAILED
+        Detection.DetectionStatus.FAILED,
+        undefined,
+        message
       )
     );
   }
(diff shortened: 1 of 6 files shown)


server/src/controllers/analytics_controller.ts
```diff
@@ -560,16 +561,35 @@ export async function getDetectionSpans(
...
+    if(!_.isEmpty(insideFailed)) {
+      return;
+    }
+
+    const intersectedFailedSpans = await Detection.getIntersectedSpans(analyticUnitId, from, to, Detection.DetectionStatus.FAILED);
+    const intersectedFailedSpansBorders = Detection.getSpanBorders(intersectedFailedSpans);
+    const filteredSpans = getNonIntersectedSpans(from, to, intersectedFailedSpansBorders);
+    _.concat(filteredNewDetectionSpans, filteredSpans);
+  });
+  Promise.all(filterFailedSpanPromises);
```
```suggestion
  await Promise.all(filterFailedSpanPromises);
```

server/src/controllers/analytics_controller.ts
```diff
@@ -560,16 +561,35 @@ export async function getDetectionSpans(
...
+    const intersectedFailedSpansBorders = Detection.getSpanBorders(intersectedFailedSpans);
+    const filteredSpans = getNonIntersectedSpans(from, to, intersectedFailedSpansBorders);
+    _.concat(filteredNewDetectionSpans, filteredSpans);
+  });
+  await Promise.all(filterFailedSpanPromises);
+
   let runningSpansPromises = [];
   let newRunningSpans: Detection.DetectionSpan[] = [];
-  runningSpansPromises = newDetectionSpans.map(async span => {
+  runningSpansPromises = filteredNewDetectionSpans.map(async span => {
```
```suggestion
  const runningSpansPromises = filteredNewDetectionSpans.map(async span => {
```

server/src/controllers/analytics_controller.ts
```diff
@@ -560,16 +561,35 @@ export async function getDetectionSpans(
...
+    }
+
+    const intersectedFailedSpans = await Detection.getIntersectedSpans(analyticUnitId, from, to, Detection.DetectionStatus.FAILED);
+    const intersectedFailedSpansBorders = Detection.getSpanBorders(intersectedFailedSpans);
+    const filteredSpans = getNonIntersectedSpans(from, to, intersectedFailedSpansBorders);
+    _.concat(filteredNewDetectionSpans, filteredSpans);
+  });
+  await Promise.all(filterFailedSpanPromises);
+
   let runningSpansPromises = [];
```
```suggestion
```

amper43:
Actual state:
Changes of span's state machine is need. Current implementation of PR works incorrect in case when panel requests range, that contain failed span. After that unit become to `FAILED` sate and can't run detections.

Issue #678, opened 2019-05-23
DetectorState class


VargBurz:
Now we use `cache` dict for detectors' state storage.
We should create a `DetectorState` class and use it. Like we did with `ModelState`.

PR #710, opened 2019-07-06
Anomaly detector bounds switch disabling to enabling #709

### LINKED ISSUE #709: Anomaly detector bounds switch: disabling to enabling
Invert analytic logic about bounds disabling for resolving this issue https://github.com/hastic/hastic-grafana-app/issues/358

amper43:
fixes #709 

### Changes
- invert logic of detections and processing
- added migration to 5 revision (disableBounds -> enabledBounds)

### Notes
For structuring of `payload` and resolving following TODO in anomaly_detector's detect method:
```
# TODO: use class for cache to avoid using string literals and Bound.TYPE.value
```
there is issue https://github.com/hastic/hastic-server/issues/678

### DIFF
--- a/analytics/analytics/detectors/anomaly_detector.py
+++ b/analytics/analytics/detectors/anomaly_detector.py
@@ -20,7 +20,7 @@ BASIC_ALPHA = 0.5
 logger = logging.getLogger('ANOMALY_DETECTOR')
 
 class Bound(Enum):
-    NONE = 'NONE'
+    ALL = 'ALL'
     UPPER = 'UPPER'
     LOWER = 'LOWER'
 
@@ -32,7 +32,7 @@ class AnomalyDetector(ProcessingDetector):
 
     def train(self, dataframe: pd.DataFrame, payload: Union[list, dict], cache: Optional[ModelCache]) -> ModelCache:
         segments = payload.get('segments')
-        disable_bound: str = payload.get('disableBound') or 'NONE'
+        enable_bounds: str = payload.get('enableBounds') or 'ALL'
         prepared_segments = []
         time_step = utils.find_interval(dataframe)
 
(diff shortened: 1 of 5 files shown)


analytics/analytics/detectors/anomaly_detector.py
```diff
@@ -79,8 +79,12 @@ def detect(self, dataframe: pd.DataFrame, cache: Optional[ModelCache]) -> Detect
         bounds[Bound.LOWER.value] = ( smoothed_data - cache['confidence'], operator.lt )
         bounds[Bound.UPPER.value] = ( smoothed_data + cache['confidence'], operator.gt )
 
-        if disable_bound != Bound.NONE.value:
-            del bounds[disable_bound]
+        if enable_bounds == Bound.LOWER.value:
```
jonyrock: `LOWER` here is like a constant. But looks like it is an object with a variable field inside.

analytics/analytics/detectors/anomaly_detector.py
```diff
@@ -32,15 +32,15 @@ def __init__(self, analytic_unit_id: AnalyticUnitId):
 
     def train(self, dataframe: pd.DataFrame, payload: Union[list, dict], cache: Optional[ModelCache]) -> ModelCache:
         segments = payload.get('segments')
-        disable_bound: str = payload.get('disableBound') or 'NONE'
+        enable_bounds: str = payload.get('enableBounds') or 'ALL'
```
jonyrock: I don't like that default value set inside train. I think it's better to send well-defined (without empty fields) config to train and raise an error is field is empty.

server/src/migrations.ts
```diff
@@ -129,6 +130,25 @@ async function addDetectorTypes() {
   await Promise.all(promises);
 }
 
+async function switchBoundsDisabling() {
+  const analyticUnits = await analyticUnitsDB.findMany({ disableBound: { $exists: true } });
+
+  const promises = analyticUnits.map(analyticUnit => {
```
jonyrock: let's make something more meaningful then just "promises"

PR #711, opened 2019-07-09
Data kit error status code 400 #300

### LINKED ISSUE #300: Data kit error: status code 400
When DataPuller tries to query data with the same `from` and `to` fields - Grafana returns error with code 400
image

amper43:
fixes #300 

### Changes
- check that current iteration in `data_puller` has `from` timestamp less than `to` and skip pulling if check fails. This need for first iteration of data pulling because now we start pull data from `now` time moment instead of `analyticUnit.lastDetectionTime`
- check that `from` timestamp is less than `to` in detection router
- in `analytic_controller` `setDetectionTime` should get right border of detection's range **without** `intersection` (payload.lastDetectionTime includes `intersection`)

### DIFF
--- a/server/src/services/data_puller.ts
+++ b/server/src/services/data_puller.ts
@@ -151,7 +151,7 @@ export class DataPuller {
         }
         const now = Date.now();
 
-        if(time === now) {
+        if(time <= now) {
           // TODO: probably we should have ability to set PULL_PERIOD_MS or got it from metric as time step between points
           return {
             columns: [],
--- a/server/src/services/data_puller.ts
+++ b/server/src/services/data_puller.ts
@@ -150,6 +150,15 @@ export class DataPuller {
           throw new Error(`Analytic unit ${analyticUnit.id} is deleted from puller`);
         }
         const now = Date.now();
+
+        if(time === now) {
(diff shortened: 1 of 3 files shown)


server/src/services/data_puller.ts
```diff
@@ -150,6 +150,15 @@ export class DataPuller {
           throw new Error(`Analytic unit ${analyticUnit.id} is deleted from puller`);
         }
         const now = Date.now();
+
+        if(time <= now) {
+          // TODO: probably we should have ability to set PULL_PERIOD_MS or got it from metric as time step between points
```
rozetko:
```suggestion
          // TODO: probably we should have ability to set PULL_PERIOD_MS or get it from metric as time step between points
```

server/src/services/data_puller.ts
```diff
@@ -150,6 +149,15 @@ export class DataPuller {
           throw new Error(`Analytic unit ${analyticUnit.id} is deleted from puller`);
         }
         const now = Date.now();
+
+        if(time >= now) {
+          // TODO: probably we should have ability to set PULL_PERIOD_MS or got it from metric as time step between points
```
jonyrock:
```suggestion
          // TODO: probably we should have ability to set PULL_PERIOD_MS or get it from metric as time step between points
```

PR #730, opened 2019-07-25
WIP: Missing piece of HSR in the end #731

VargBurz:
HSR  is cut in the end of the chart.

Will be fixed after #733 

## Changes
- Add `timestep` (difference between first 2 dataset timestamps)
- Add second HSR query for `[to; to + timestep]` range

## PATTERN BEFORE:
image
## PATTERN AFTER:
image

## ANOMALY BEFORE:
image
## ANOMALY AFTER:
image

### DIFF
--- a/server/src/controllers/analytics_controller.ts
+++ b/server/src/controllers/analytics_controller.ts
@@ -629,9 +629,14 @@ export async function getHSR(
   upperBound?: TableTimeSeries
 }> {
   try {
+    //TODO: Use aggregation time instead of 2nd query when it's supported by grafana-datasource-kit
+    let timestep = 86400000;
     const grafanaUrl = getGrafanaUrl(analyticUnit.grafanaUrl);
-    const data = await queryByMetric(analyticUnit.metric, grafanaUrl, from, to, HASTIC_API_KEY);
-
+    let data = await queryByMetric(analyticUnit.metric, grafanaUrl, from, to, HASTIC_API_KEY);
+    if (data.values != undefined && data.values.length > 1) {
+      timestep = data.values[1][0] - data.values[0][0];
+    }
+    data = await queryByMetric(analyticUnit.metric, grafanaUrl, from, to + timestep, HASTIC_API_KEY);
     if(analyticUnit.detectorType !== AnalyticUnit.DetectorType.ANOMALY) {
       return { hsr: data };
     }
--- a/server/src/controllers/analytics_controller.ts
+++ b/server/src/controllers/analytics_controller.ts
@@ -630,7 +630,7 @@ export async function getHSR(
 }> {
   try {
     //TODO: Use aggregation time instead of 2nd query when it's supported by grafana-datasource-kit
-    let timestep = 86400000;
+    let timestep = 0;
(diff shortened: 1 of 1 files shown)


server/src/controllers/analytics_controller.ts
```diff
@@ -629,9 +629,14 @@ export async function getHSR(
   upperBound?: TableTimeSeries
 }> {
   try {
+    //TODO: Use aggregation time instead of 2nd query when it's supported by grafana-datasource-kit
+    let timestep = 86400000;
```
rozetko: Maybe it's better to use 0 as default timestep?

VargBurz: Default timestep can help if this variable cannot be calculated.

jonyrock:
@VargBurz please add a screenshot: BEFORE and AFTER into the description

server/src/controllers/analytics_controller.ts
VargBurz: @rozetko update variable

server/src/controllers/analytics_controller.ts
```diff
@@ -629,9 +629,14 @@ export async function getHSR(
...
+    //TODO: Use aggregation time instead of 2nd query when it's supported by grafana-datasource-kit
+    let timestep = 0;
     const grafanaUrl = getGrafanaUrl(analyticUnit.grafanaUrl);
-    const data = await queryByMetric(analyticUnit.metric, grafanaUrl, from, to, HASTIC_API_KEY);
-
+    let data = await queryByMetric(analyticUnit.metric, grafanaUrl, from, to, HASTIC_API_KEY);
+    if (data.values != undefined && data.values.length > 1) {
+      timestep = data.values[1][0] - data.values[0][0];
+    }
+    data = await queryByMetric(analyticUnit.metric, grafanaUrl, from, to + timestep, HASTIC_API_KEY);
```
rozetko: We can do this query only inside the `if`. Otherwise, we do similar queries twice.
Also, query can be done on range `[to; to+timestep]` and appended to `data`

server/src/controllers/analytics_controller.ts
```diff
@@ -629,9 +629,14 @@ export async function getHSR(
...
   try {
+    //TODO: Use aggregation time instead of 2nd query when it's supported by grafana-datasource-kit
+    let timestep = 0;
     const grafanaUrl = getGrafanaUrl(analyticUnit.grafanaUrl);
-    const data = await queryByMetric(analyticUnit.metric, grafanaUrl, from, to, HASTIC_API_KEY);
-
+    let data = await queryByMetric(analyticUnit.metric, grafanaUrl, from, to, HASTIC_API_KEY);
+    if (data.values != undefined && data.values.length > 1) {
+      timestep = data.values[1][0] - data.values[0][0];
+      data = await queryByMetric(analyticUnit.metric, grafanaUrl, from, to + timestep, HASTIC_API_KEY);
```
jonyrock: so you make a query at line 635 and the you make another one here, right? 
what if make a proper if/else code flow?

PR #740, opened 2019-08-02
Error cannot read property data of null #739

### LINKED ISSUE #739: Error: Cannot read property 'data' of null
We should check that `data` is not `null` here
https://github.com/hastic/hastic-server/blob/master/server/src/models/analytic_unit_cache_model.ts#L36

`data` contains cache, it is a responce from DB
Data will be null if analytic unit hasn't got though "learning" stage

amper43:
fixes #739 

## Problem

in `getIntersection` there isn't a check that `this.data` is null. Data will be null when analytic unit haven't got though "learning" stage. It happens when user creates an analytic unit and goes to inspect mode.

## Changes
- added check for `cache.data` === `null` case
- `getHSR`: check threshold's fields for case when threshold needs learning

### DIFF
--- a/server/src/controllers/analytics_controller.ts
+++ b/server/src/controllers/analytics_controller.ts
@@ -3,6 +3,7 @@ import { AnalyticsTask, AnalyticsTaskType, AnalyticsTaskId } from '../models/ana
 import * as AnalyticUnitCache from '../models/analytic_unit_cache_model';
 import * as Segment from '../models/segment_model';
 import * as AnalyticUnit from '../models/analytic_units';
+import { ThresholdAnalyticUnit, AnomalyAnalyticUnit} from '../models/analytic_units';
 import * as Detection from '../models/detection_model';
 import { AnalyticsService } from '../services/analytics_service';
 import { AlertService } from '../services/alert_service';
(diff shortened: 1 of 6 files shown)


jonyrock:
> null `this.data` case? 

Sorry, I don't understand. Please explain which case

amper43:
> > null `this.data` case?
> 
> Sorry, I don't understand. Please explain which case

description updated

server/src/controllers/analytics_controller.ts
```diff
@@ -659,9 +659,17 @@ export async function getHSR(
 
     let cache = await AnalyticUnitCache.findById(analyticUnit.id);
     if(
-      cache === null ||
-      cache.data.alpha !== (analyticUnit as AnalyticUnit.AnomalyAnalyticUnit).alpha ||
-      cache.data.confidence !== (analyticUnit as AnalyticUnit.AnomalyAnalyticUnit).confidence
+        cache === null ||
```
jonyrock: it is the biggest `if` statement I saw in my entire software engineering career. 

Please make a function for this.

server/src/models/analytic_unit_cache_model.ts
```diff
@@ -33,7 +33,7 @@ export class AnalyticUnitCache {
   }
 
   public getIntersection(): number {
-    if(this.data.windowSize !== undefined) {
+    if(this.data !== null && this.data.windowSize !== undefined) {
```
jonyrock: can `this.data` be `undefined`?   
```
undefined !== null // true
```

server/src/controllers/analytics_controller.ts
```diff
@@ -658,11 +658,8 @@ export async function getHSR(
     }
 
     let cache = await AnalyticUnitCache.findById(analyticUnit.id);
-    if(
-      cache === null ||
-      cache.data.alpha !== (analyticUnit as AnalyticUnit.AnomalyAnalyticUnit).alpha ||
-      cache.data.confidence !== (analyticUnit as AnalyticUnit.AnomalyAnalyticUnit).confidence
-    ) {
+    const analyticFields = analyticUnit.getAnalyticFields();
+    if(cache === null || !_.every(_.keys(analyticFields).map(k => analyticFields[k] === cache.data[k]))) {
```
jonyrock:
```suggestion
    let someSemanticVarName = _.every(_.keys(analyticFields).map(k => analyticFields[k] === cache.data[k]));
    if(cache === null || !someMeamingfullVariableName) {
```

jonyrock: I see that you use `cache.data[k]` in `someSemanticVarName`
Well, you better write a bool function which resolves what you want. So it's better
```js
analyticUnitFieldsEqualsToCache(analyticUnit: AnalyticUnit, cache?: CacheModel): boolean {
  if(cache === null) {
    return false;
  }
  // it's better to use https://lodash.com/docs/4.17.15#isEqual here
  var analyticFields = analyticUnit.getAnalyticFields();
  return _.every(_.keys(analyticFields).map(k => analyticFields[k] === cache.data[k])
}
```

server/src/models/analytic_units/analytic_unit_model.ts
```diff
@@ -78,4 +77,6 @@ export abstract class AnalyticUnit {
     };
   }
 
+  public getAnalyticFields(): any {};
```
jonyrock: just `any` , what if `any[]` ?

PR #755, opened 2019-08-17
WIP: Utils.spans.cutSpanWithSpans more tests

jonyrock:
Just to be sure that the function `Utils.spans.cutSpanWithSpans` works as it should, few more tests added: It will help in future if we want to improve the performance.

This PR changes the [behavior](https://github.com/hastic/hastic-server/pull/755#issuecomment-522265390) of `cutSpanWithSpans` and renames it to `cutSegmentWithSegments`

### One-point cut
If we cut `[1, 10]` with `[5, 5]` , we should get spans `[1, 4]` and `[6, 10]`

### Infinite spans
If we cut `[0, Infinity]` with `[5, 10]`, we should get `[0, 4]` and `[11, Infinity]`

If we cut `[1, 6]` with `[0, Infinity]`, we should get `[]`

If we cut `[0, 6]` with `[1, Infinity]`, we should get `[0, 0]`

If we cut `[-Infinity, Infinity]` with `[-Infinity, Infinity]` we should get `[]`

### Overlapping cuts

There are many combinations where cuts include each other or could be sorted wrong.
Many new combinations added. Including infinite case.

### DIFF
--- a/server/spec/utils/spans.jest.ts
+++ b/server/spec/utils/spans.jest.ts
@@ -13,7 +13,7 @@ function cutSpan(from: number, to: number, cuts: [number, number][]): [number, n
 describe('cutSpanWithSpans', function() {
 
   it('should find spans in simple non-intersected borders', function() {
-    let cutSpans = [[3, 5], [6, 8], [10, 20]] as [number, number][];
+    let cutSpans = [[3, 4], [6, 8], [11, 20]] as [number, number][];
 
     expect(cutSpan(4, 11, cutSpans)).toEqual([[5, 6], [8, 10]]);
     expect(cutSpan(5, 11, cutSpans)).toEqual([[5, 6], [8, 10]]);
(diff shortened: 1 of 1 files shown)


jonyrock:
@amper43 @rozetko 

In first version of this PR, I said:
> ### One-point cut
> If we cut [1, 10] with [5, 5] , we should get spans [1, 5] and [5, 10]

but is really true behavior? In theory we can get such case when we run detector for one point for threshold.

jonyrock:
I believe that `Span` should represent a discrete structure, which is a [countable set](https://en.wikipedia.org/wiki/Countable_set), [integer segment](https://en.wikipedia.org/wiki/Line_segment) , subset of [Z](https://en.wikipedia.org/wiki/Integer)

So when we want to get cut `[x, x]` from `[a, b]` where `a < x < b`, then we should return split `[a, x - 1]` and `[x + 1, b]`

jonyrock:
Also, I propose to stop using word "Span" without clear math definition. So I an going to rename `Utils.spans.cutSpanWithSpans` accordingly.

jonyrock:
image

jonyrock:
All thread compiled to https://github.com/hastic/hastic-server/pull/756

PR #756, opened 2019-08-17
Utils SpansCut to SegmentsCut, IntegerSegment and IntegerSegmentSet

jonyrock:
Just to be sure that the function `Utils.spans.cutSpanWithSpans` works as it should, few more tests added: It will help in future if we want to improve the performance.

I believe that `Span` should represent a discrete structure, which is a [countable set](https://en.wikipedia.org/wiki/Countable_set), [integer segment](https://en.wikipedia.org/wiki/Line_segment) , subset of [Z](https://en.wikipedia.org/wiki/Integer)
Also, I propose to stop using word "Span" without clear math definition. So I an going to rename `Utils.spans.cutSpanWithSpans` accordingly. 

image

If we cut  `[z, z]` and `[x, y]` from `[a, b]` , then we get `[[a, z -  1], [z + 1, x - 1]]`

So when we want to get cut `[x, x]` from `[a, b]` where `a < x < b`, then we should return split `[a, x - 1]` and `[x + 1, b]`

This PR changes the [behavior](https://github.com/hastic/hastic-server/pull/755#issuecomment-522265390) of `cutSpanWithSpans` and renames it to `cutSegmentWithSegments`

## Changes
* New class `IntegerSegment` -- it is like segment, but integer
* New class `IntegerSegmentsSet` -- can eat IntegerSegments and has set operations. For now implemented operations necessary for `cutSegmentWithSegments`

### One-point cut
If we cut `[1, 10]` with `[5, 5]` , we should get spans `[1, 4]` and `[6, 10]`

### Infinite spans
If we cut `[0, Infinity]` with `[5, 10]`, we should get `[0, 4]` and `[11, Infinity]`

If we cut `[1, 6]` with `[0, Infinity]`, we should get `[]`

If we cut `[0, 6]` with `[1, Infinity]`, we should get `[0, 0]`

If we cut `[-Infinity, Infinity]` with `[-Infinity, Infinity]` we should get `[]`

### Overlapping cuts

There are many combinations where cuts include each other or could be sorted wrong.
Many new combinations added. Including infinite case.


server/src/utils/segments.ts
```diff
@@ -0,0 +1,186 @@
...
+ * @returns array of segments remain after cut
+ */
+export function cutSegmentWithSegments(inputSegment: Segment, cutSegments: Segment[]): Segment[] {
+  let setA = new IntegerSegmentsSet([new IntegerSegment(inputSegment.from, inputSegment.to)]);
+  let setB = new IntegerSegmentsSet(cutSegments.map(
+    s => new IntegerSegment(s.from, s.to)
+  ));
+  let setResult = setA.sub(setB);
+  return setResult.segments.map(s => ({ from: s.from, to: s.to }));
+  return []
```
rozetko:
```suggestion
```

server/src/utils/segments.ts
```diff
@@ -0,0 +1,186 @@
...
+    let inversed = other.inversed();
+    return this.intersected(inversed);
+  }
+
+}
+
+// TODO: move from utils and use generator
+/**
+ *
+ * @param inputSegment a big span which we will cut
```
rozetko:
```suggestion
 * @param inputSegment a big segment which we will cut
```

server/src/utils/segments.ts
```diff
@@ -0,0 +1,186 @@
...
+ * @param cutSegments spans which to cut the inputSpan. Spans can overlay.
```
rozetko:
```suggestion
 * @param cutSegments segments to cut the inputSegment. Segments can overlay.
```

server/src/utils/segments.ts
```diff
@@ -0,0 +1,186 @@
...
+          push(prev.to + 1, s.from - 1);
+        }
+        return s;
+      }, null);
+      push(_.last(this._segments).to + 1, Infinity);
+    }
+    return new IntegerSegmentsSet(invertedSegments, true);
+  }
+
+  intersected(other: IntegerSegmentsSet): IntegerSegmentsSet {
```
rozetko: Maybe it can be `intersect` or `getIntersection`?

server/spec/utils/segments.jest.ts
```diff
@@ -0,0 +1,204 @@
...
+  it('should sort segments', function() {
+    expect(ISS([[10, 15], [5, 8]]).segments).toEqual([IS(5, 8), IS(10, 15)]);
+    expect(ISS([[10, 15], [-Infinity, 8]]).segments).toEqual([IS(-Infinity, 8), IS(10, 15)]);
+    expect(ISS([[10, Infinity], [-Infinity, 8]]).segments).toEqual([IS(-Infinity, 8), IS(10, Infinity)]);
+  });
+
+  it('should merge segments', function() {
+    expect(ISS([[5, 10], [7, 20]]).segments).toEqual([IS(5, 20)]);
+    expect(ISS([[5, 10], [10, 20]]).segments).toEqual([IS(5, 20)]);
+    expect(ISS([[5, 10], [11, 20]]).segments).toEqual([IS(5, 20)]);
```
rozetko: Why do these segments get merged?

jonyrock: @rozetko I will add a comment into the code

server/src/utils/segments.ts
jonyrock: @rozetko what if `intersect` ?

rozetko: @jonyrock would be much better :)

Issue #765, opened 2019-08-25
tsconfig constrains


jonyrock:
Let's enforce compilation constrains in `.tsconfig` from here: https://www.typescriptlang.org/docs/handbook/compiler-options.html

PR #767, opened 2019-08-26
Tsconfig strict mode part1

jonyrock:
Fixes https://github.com/hastic/hastic-server/issues/765
 
## Changes
* `strict` mode to `.tsconfig` more info: https://www.typescriptlang.org/docs/handbook/compiler-options.html. This PR is "part1" so actual strict mode is disabled because otherwise project does not compile.
* refactoring of `data_service.ts`: nedb object inserted to database manipulation functions as first param instead of `Collection` -- an Enum for all possible collections.
* `mergeAndInsertSegments` : querying of `AnalyticUnit.findById(analyticUnitId)` only once 

## Found mistakes in code

`services/data_service.ts`
```js
let dbRemoveMany = (...): Promise<number> {
  if(isEmptyArray(query)) {
      return Promise.resolve([]); // it should return 0 here
  }
  // ...
```


server/package.json
```diff
@@ -21,7 +21,7 @@
   "homepage": "https://github.com/hastic/hastic-server#readme",
   "dependencies": {},
   "devDependencies": {
-    "@types/jest": "^23.1.1",
+    "@types/jest": "^23.3.14",
```
rozetko: why is it changed? `jest` version is still `23.1.1`

PR #769, opened 2019-08-28
rm try in some test + typos fix + whitespaces fix

jonyrock:
This PR changes constructions in tests like

```python

def test():
    try:
        # ...
    catch():
        self.fail()
```
There is no need for this. 

Also small typos and whitespaces fixes


jonyrock:
@amper43 I hope you can help me there with CI

Issue #570, opened 2019-04-15
Hastic persistence database


jonyrock:
Hastic has to save it's artifacts to a database. We chose [nedb](https://github.com/louischatriot/nedb/) as a small test storage which we can easy add as a requirement to run Hastic. Obviously, we can't scale this approach much, so we need a "legit" database.

I propose to support all databases which we have in [grafana-datasource-kit](https://github.com/CorpGlory/grafana-datasource-kit)

Let's start with [influxdb](https://www.influxdata.com/)

jonyrock:
Because we use nedb, let's take MongoDB as the first "real database support"

amper43:
subtasks:
- [ ] add config fields
- [ ] fix data service to work with external data base
- [ ] add management of mongodb lifecycle in production mode
- [ ] add mongodb to docker-compose

PR #771, opened 2019-09-04
Mongodb persistence #570 part 1

amper43:
First iteration of MongoDb integration

## Changes
- added related config fields, including `HASTIC_EXTERNAL_DB`  that enable using MongoDB
- added authentification to data base
- added mapping to collections in db
- added synchronious way to connect to db. autoamtically reconnect enabled.

## Current state
previous way with NeDb works. if user enable `HASTIC_EXTERNAL_DB` hastic will crash.
In second iteration need make fabric method that return according implementation because NeDb's interface supports only callbacks not promises, but MongoDb interface supports promises.

### DIFF
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -12,9 +12,14 @@ let configExists = fs.existsSync(configFile);
 export const ANALYTICS_PATH = path.join(__dirname, '../../analytics');
 
 export const HASTIC_DB_IN_MEMORY = getConfigField('HASTIC_IN_MEMORY_PERSISTANCE', false);
+export const HASTIC_EXTERNAL_DB = getConfigField('HASTIC_EXTERNAL_DB', false);
+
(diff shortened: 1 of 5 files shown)


server/src/services/data_service.ts
```diff
@@ -13,6 +15,15 @@ export enum Collection {
   DB_META
 };
 
+export const NamesCollection = {
```
rozetko: let's use `Map` here

server/src/config.ts
```diff
@@ -114,3 +129,18 @@ function createZMQConnectionString() {
...
 }
+
+function getDbConfig(connectionStr: string): DataBaseConfig {
+  // TODO: refactor
+  const [USER, PASSWORD] = connectionStr.split('@')[0].split(':');
+  const [DB_NAME, ...urlParts] = connectionStr.split('@')[1].split('/').reverse();
+  const URL = urlParts.reverse().join('/');
+
+  const config = {
+    USER,
```
rozetko: why do we need caps here?

PR #773, opened 2019-09-06
Learning is started even if analytics is down #726

### LINKED ISSUE #726: Learning is started even if analytics is down
Analytic unit first gets `LEARNING` status and task is sent to analytics only after that
We should set `LEARNING` status **after** sending the task

VargBurz:
Closes #726 

## Changes:

- Set status `LEARNING` *after* successful check that analytics is ready

### DIFF
--- a/server/src/controllers/analytics_controller.ts
+++ b/server/src/controllers/analytics_controller.ts
@@ -327,7 +327,7 @@ export async function runLearning(id: AnalyticUnit.AnalyticUnitId, from?: number
     }
     let result = await runTask(task);
     if(result.status === AnalyticUnit.AnalyticUnitStatus.FAILED) {
-      throw new Error(`Analytics failed with an error: ${result.error}`);
+      throw new Error(`Task for analytics failed with an error: ${result.error}`);
     }
     // TODO: rename SUCCESS to something better
     await AnalyticUnit.setStatus(id, AnalyticUnit.AnalyticUnitStatus.SUCCESS);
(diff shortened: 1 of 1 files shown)


server/src/controllers/analytics_controller.ts
```diff
@@ -321,12 +321,12 @@ export async function runLearning(id: AnalyticUnit.AnalyticUnitId, from?: number
     let task = new AnalyticsTask(
       id, AnalyticsTaskType.LEARN, taskPayload
     );
-    AnalyticUnit.setStatus(id, AnalyticUnit.AnalyticUnitStatus.LEARNING);
     console.log(`run ${task.type} task, id:${id}`);
     let result = await runTask(task);
     if(result.status !== AnalyticUnit.AnalyticUnitStatus.SUCCESS) {
       throw new Error(result.error);
```
jonyrock: @VargBurz @rozetko 
if we changing this part of code, let's wrap this `Error` in something better than `throw new Error(result.error)`

server/src/controllers/analytics_controller.ts
```diff
@@ -324,7 +324,7 @@ export async function runLearning(id: AnalyticUnit.AnalyticUnitId, from?: number
     console.log(`run ${task.type} task, id:${id}`);
     let result = await runTask(task);
     if(result.status === AnalyticUnit.AnalyticUnitStatus.FAILED) {
-      throw new Error(result.error);
+      throw new Error(`Analytics failed with an error: ${result.error}`);
```
jonyrock: I believe that not entire analytics failed, but a particular task, right?

PR #778, opened 2019-09-25
Error: $or must be an array

VargBurz:
### Problem
There is an error in the detection process of Anomaly or Pattern Detector with MongoDB:
```
Error: $or must be an array
    at Object.<anonymous> (/mnt/c/Users/VargB/git/hastic-server/server/dist/server-dev.js:528:19)
    at Generator.next (<anonymous>)
    at fulfilled (/mnt/c/Users/VargB/git/hastic-server/server/dist/server-dev.js:232:32)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

image

### Reason

MongoDB and nedb have different `$or` field syntax in `find` query.

### Changes

- change `findMany` query syntax from `$or: { labeled: true, deleted: true}` to` $or [{labeled: true}, {deleted: true}]` for MongoDB

### DIFF
--- a/server/src/services/data_layer/mongodb.ts
+++ b/server/src/services/data_layer/mongodb.ts
@@ -121,7 +121,7 @@ export class MongoDbQueryWrapper implements DbQueryWrapper {
   }
 }
 
-function convertQueryToMongoFormat(query: any): object {
+function convertQueryToMongoFormat(query: { $or?: object, $and?: object }): object {
   if(query.$or !== undefined) {
     query.$or = convertQueryFieldToMongoFormat(query.$or);
   }
--- a/server/src/services/data_layer/mongodb.ts
+++ b/server/src/services/data_layer/mongodb.ts
@@ -1,7 +1,7 @@
-import { DbQueryWrapper } from './basedb';
+import { DbQueryWrapper, QueryExecutionError } from './basedb';
 
 import { Collection, FilterQuery, ObjectID } from 'mongodb';
-import { wrapIdToMongoDbQuery, wrapIdsToMongoDbQuery, isEmptyArray } from './utils';
+import { wrapIdToMongoDbQuery, wrapIdsToMongoDbQuery, isEmptyArray, useMongoSyntax } from './utils';
 
 import * as _ from 'lodash';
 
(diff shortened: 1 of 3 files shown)


jonyrock:
But I think we need to throw a nice exception if we get another error during the processing of nedb / mongo query. Something like `QueryExecutionError`

server/src/services/data_layer/mongodb.ts
```diff
@@ -79,20 +79,32 @@ export class MongoDbQueryWrapper implements DbQueryWrapper {
...
+      const docs = await collection.find(query).sort(sortQuery).toArray();
+      // TODO: move to utils
+      docs.forEach(doc => {
+        if (doc !== null) {
+          doc._id = doc._id.toString();
+        }
+      });
+      return docs;
+    } catch(error) {
+      console.error(`Can't find query in collection: ${collection.namespace}`);
```
rozetko: what does "find query" mean?

VargBurz: get query result?

rozetko: sounds better :)
I'd also add query itself to this message

server/src/services/data_layer/mongodb.ts
```diff
@@ -79,20 +79,32 @@ export class MongoDbQueryWrapper implements DbQueryWrapper {
     return doc;
   }
 
-  async dbFindMany(collection: Collection, query: string[] | object, sortQuery: object = {}): Promise<any[]> {
+  async dbFindMany(collection: Collection, query: any, sortQuery: object = {}): Promise<any[]> {
```
amper43: why type of `query` was changed? lets keep type `string[] | object` or add required type.

server/src/services/data_layer/mongodb.ts
```diff
@@ -79,20 +79,32 @@ export class MongoDbQueryWrapper implements DbQueryWrapper {
...
-    // TODO: move to utils
-    docs.forEach(doc => {
-      if(doc !== null) {
-        doc._id = doc._id.toString();
-      }
-    });
-    return docs;
+    try{
+      const docs = await collection.find(query).sort(sortQuery).toArray();
+      // TODO: move to utils
```
amper43: lets remove comment if it isn't actual

server/src/services/data_layer/mongodb.ts
VargBurz: I need to use a certain field.
But I can't affix a certain type. because of different collections.

server/src/services/data_layer/mongodb.ts
```diff
@@ -84,15 +84,20 @@ export class MongoDbQueryWrapper implements DbQueryWrapper {
...
+      const docs = await collection.find(query).sort(sortQuery).toArray();
+      docs.forEach(doc => {
+        if (doc !== null) {
+          doc._id = doc._id.toString();
+        }
+      });
+      return docs;
+    } catch(error) {
+      console.error(`Can't get query result for query ${JSON.stringify(query)} in collection: ${collection.namespace}`);
+      throw new QueryExecutionError(`mongo query problem: ${error.message}`);
```
rozetko:
```suggestion
      throw new QueryExecutionError(`MongoDB query error: ${error.message}`);
```

server/src/services/data_layer/mongodb.ts
```diff
@@ -84,15 +84,20 @@ export class MongoDbQueryWrapper implements DbQueryWrapper {
...
     query = wrapIdsToMongoDbQuery(query);
-    const docs = await collection.find(query).sort(sortQuery).toArray();
-    // TODO: move to utils
-    docs.forEach(doc => {
-      if(doc !== null) {
-        doc._id = doc._id.toString();
-      }
-    });
-    return docs;
+    try{
```
rozetko:
```suggestion
    try {
```

PR #781, opened 2019-10-01
Notifications timezone #649

### LINKED ISSUE #649: Notifications timezone
It is possible want to receive notifications in your local time. I believe that it should be the same as dashboard. User would set his/her preferred time.

VargBurz:
Closes #649 

Notification message has wrong time info due to the time difference.

## Changes:

- add optinal field in config "TIMEZONE_UTC_OFFSET", where you can specify any timezone for getting notifications.
- if there is no field, server will be used default value equals your local timezone offset.
- If the field exists, but has wrong format, there will be an error and server will not start.
- When notification message creates, `from` and `to` timestamps replace for utc time and than add utc_offset.


server/src/config.ts
```diff
@@ -50,6 +51,7 @@ export const HASTIC_WEBHOOK_URL = getConfigField('HASTIC_WEBHOOK_URL', null);
 export const HASTIC_WEBHOOK_TYPE = getConfigField('HASTIC_WEBHOOK_TYPE', 'application/json');
 export const HASTIC_WEBHOOK_SECRET = getConfigField('HASTIC_WEBHOOK_SECRET', null);
 export const HASTIC_WEBHOOK_IMAGE_ENABLED = getConfigField('HASTIC_WEBHOOK_IMAGE', false);
+export const TIMEZONE_UTC_OFFSET = getTimeZoneOffset('TIMEZONE_UTC_OFFSET');
```
amper43: let's move ''TIMEZONE_UTC_OFFSET" literal to `getTimeZoneOffset` method

server/src/config.ts
```diff
@@ -145,3 +147,22 @@ function getDbConfig(connectionStr: string): DBConfig {
...
+    return parseTimeZone(configTimeZone);
+  } else {
+    const serverUtcOffset = moment().utcOffset();
+    return serverUtcOffset;
+  }
+}
+
+export function parseTimeZone(timeZone: string): Number {
+  const time = _.split(timeZone, ':');
+  let minutsOffset = Math.abs(Number(time[0])) * 60 + Number(time[1]);
```
amper43: `minutesOffset`

let's add name for number `60`

server/src/config.ts
```diff
@@ -145,3 +147,22 @@ function getDbConfig(connectionStr: string): DBConfig {
...
+  if(configTimeZone !== null) {
+    return parseTimeZone(configTimeZone);
+  } else {
+    const serverUtcOffset = moment().utcOffset();
+    return serverUtcOffset;
+  }
+}
+
+export function parseTimeZone(timeZone: string): Number {
+  const time = _.split(timeZone, ':');
```
amper43: let's check `timeZone` have correct format

server/src/config.ts
```diff
@@ -145,3 +148,27 @@ function getDbConfig(connectionStr: string): DBConfig {
...
+  let configTimeZone = getConfigField('TIMEZONE_UTC_OFFSET', null);
+  if(configTimeZone !== null) {
+    return parseTimeZone(configTimeZone);
+  } else {
+    const serverUtcOffset = moment().utcOffset();
+    return serverUtcOffset;
+  }
+}
+
+export function parseTimeZone(timeZone: string): number {
```
jonyrock: Not sure it's a right place for this function. Maybe utils ?

server/src/services/alert_service.ts
```diff
@@ -2,12 +2,19 @@ import { sendNotification, MetaInfo, AnalyticMeta, WebhookType, Notification } f
...
 import { availableReporter } from '../utils/reporter';
-import { ORG_ID, HASTIC_API_KEY, HASTIC_WEBHOOK_IMAGE_ENABLED } from '../config';
+import { ORG_ID, HASTIC_API_KEY, HASTIC_WEBHOOK_IMAGE_ENABLED, TIMEZONE_UTC_OFFSET } from '../config';
 
 import axios from 'axios';
 import * as _ from 'lodash';
+import * as moment from 'moment';
 
 
+export function toTimeZone(time: moment.MomentInput): string {
```
jonyrock: move it to utils

PR #786, opened 2019-10-07
Add timezone info to notification message

VargBurz:
It is fix for #782.

## Changes
 - add timezone info to notification message

## Before
```
...
Starts at: Fri Oct 04 2019 19:27:46
...
```
## After
```
...
Starts at: Fri Oct 04 2019 19:27:46 UTC+03:00
...
```

### DIFF
--- a/server/src/utils/time.ts
+++ b/server/src/utils/time.ts
@@ -4,6 +4,7 @@ import * as _ from 'lodash';
 import * as moment from 'moment';
 
 const MINUTES_IN_HOUR = 60;
+const TIME_FORMAT = 'ddd MMM DD YYYY HH:mm:ss Z';
 
 export function parseTimeZone(timeZone: string): number {
   timeZone = timeZone.replace(/['|"]/g, '');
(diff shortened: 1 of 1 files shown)


server/src/utils/time.ts
```diff
@@ -23,5 +23,19 @@ export function parseTimeZone(timeZone: string): number {
 export function toTimeZone(time: moment.MomentInput): string {
   const utcTime = moment(time).utc();
   const timeWithOffset = utcTime.utcOffset(TIMEZONE_UTC_OFFSET);
-  return timeWithOffset.format('ddd MMM DD YYYY HH:mm:ss');
+  const timeInfo = timeWithOffset.format('ddd MMM DD YYYY HH:mm:ss') + `(UTC ${getTimeFromMins(TIMEZONE_UTC_OFFSET)})`;
+  return timeInfo;
 }
+
+function getTimeFromMins(mins: number): string {
```
rozetko: We can add `Z` to format-string instead of this function
https://momentjs.com/docs/#/displaying/format/

PR #791, opened 2019-10-15
"Error: Need at least 1 labeled segment" when labeling only negative segment #790

### LINKED ISSUE #790: "Error: Need at least 1 labeled segment" when labeling only negative segment
User can label only negative segment and don't label any positive ones
"Error: Need at least 1 labeled segment" would raise in this case
Maybe we should not allow labeling saving without positive segments?

Related task: hastic/hastic-grafana-app#254

VargBurz:
Closes #790 

### Changes
- If there is no segments after labeling, server throws an error: `"Error: Need at least 1 labeled segment"`
- If there is only negative segments, pattern detector throws an error: `"{analytic_unit_id}has no positive labeled segments. Pattern detector needs at least 1 positive labeled segment" `
- Segments are converted from dict to class in `analytic_unit_manager.py`, not in `model.py`
- Add 2 fields to the `Segment` class: `labeled` and `deleted`
- Use `JSONClass` decorator for `Segment` class

### DIFF
--- a/analytics/analytics/analytic_types/segment.py
+++ b/analytics/analytics/analytic_types/segment.py
@@ -5,16 +5,20 @@ class Segment:
     Used for segment manipulation instead of { 'from': ..., 'to': ... } dict
     '''
 
-    def __init__(self, from_timestamp: int, to_timestamp: int, message: str = None):
+    def __init__(self, from_timestamp: int, to_timestamp: int, labeled: bool = None, deleted: bool = None, message: str = None):
         if to_timestamp < from_timestamp:
             raise ValueError(f'Can`t create segment with to < from: {to_timestamp} < {from_timestamp}')
         self.from_timestamp = from_timestamp
         self.to_timestamp = to_timestamp
+        self.labeled = labeled
+        self.deleted = deleted
         self.message = message
 
     def to_json(self):
         return {
             'from': self.from_timestamp,
             'to': self.to_timestamp,
+            'labeled': self.labeled,
+            'deleted': self.deleted,
             'message': self.message
         }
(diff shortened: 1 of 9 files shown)


analytics/analytics/analytic_types/segment.py
```diff
@@ -5,16 +5,20 @@ class Segment:
     Used for segment manipulation instead of { 'from': ..., 'to': ... } dict
     '''
 
-    def __init__(self, from_timestamp: int, to_timestamp: int, message: str = None):
+    def __init__(self, from_timestamp: int, to_timestamp: int, labeled: bool = None, deleted: bool = None, message: str = None):
```
rozetko: `labeled` and `deleted` should be `Optional[bool]` as you set it to `None` by default

analytics/analytics/detectors/pattern_detector.py
```diff
@@ -134,3 +140,8 @@ def get_window_size(self, cache: Optional[ModelCache]) -> int:
         # TODO: windowSize -> window_size
         return cache.get('windowSize', self.DEFAULT_WINDOW_SIZE)
 
+    def check_labeled_segments(self, segments: List[Segment]) -> bool:
```
rozetko:
```suggestion
    def contains_labeled_segments(self, segments: List[Segment]) -> bool:
```

analytics/analytics/analytic_unit_manager.py
```diff
@@ -66,7 +67,9 @@ def __ensure_worker(
             return res
         elif task['type'] == 'LEARN':
             if 'segments' in payload:
-                return await worker.do_train(payload['segments'], data, payload['cache'])
+                segments = payload['segments']
+                segments = list(map(lambda segment: Segment.from_json(segment), segments))
```
rozetko: segments = [Segment.from_json(segment) for segment in segments]

Issue #830, opened 2020-01-22
Export / import analytic units


rozetko:
It should be possible to export / import analytic units to make a migration between Hastic instances easier

PR #833, opened 2020-01-28
Export analytic units

rozetko:
This PR partially implements #830: adds an endpoint for exporting analytic units.

## Changes
- new `GET /panels/template` endpoint which returns:
  - analytic unit templates (analytic units with some fields changed to variables)
  - analytic unit caches
  - segments
  - detection spans
- new types for serialized analytic units
- `findMany` method for AnalyticUnitCache
- `findByAnalyticUnitIds` method for DetectionSpans and Segments

## Notes
We use TypeScript v2.6.8 which lacks ["Omit" type](https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/#the-omit-helper-type) from TS 3.5.1. This PR adds this type to `server/src/types.ts`, we can remove it when we'll update TS.

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -1,7 +1,8 @@
 import { router as analyticUnitsRouter } from './routes/analytic_units_router';
 import { router as segmentsRouter } from './routes/segments_router';
 import { router as dataRouter } from './routes/data_router';
-import { router as detectionsRouter }  from './routes/detections_router';
+import { router as detectionsRouter } from './routes/detections_router';
+import { router as panelRouter } from './routes/panel_router';
 
 import * as AnalyticsController from './controllers/analytics_controller';
 
@@ -57,6 +58,7 @@ async function init() {
   rootRouter.use('/segments', segmentsRouter.routes(), segmentsRouter.allowedMethods());
   rootRouter.use('/query', dataRouter.routes(), dataRouter.allowedMethods());
   rootRouter.use('/detections', detectionsRouter.routes(), detectionsRouter.allowedMethods());
+  rootRouter.use('/panel', panelRouter.routes(), panelRouter.allowedMethods());
 
   rootRouter.get('/', async (ctx) => {
     const activeWebhooks = await AnalyticsController.getActiveWebhooks();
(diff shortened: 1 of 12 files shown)


server/src/routes/panel_router.ts
```diff
@@ -0,0 +1,18 @@
...
+    throw new Error('Cannot export analytic units with undefined panelId');
+  }
+
+  const json = await exportPanel(panelId);
+  ctx.response.body = json;
+}
+
+export var router = new Router();
+
+router.get('/template', getPanelTemplate);
```
jonyrock-back:
```suggestion
router.get('/panel-template', getPanelTemplate);
```

rozetko: @jonyrock-back
Don't think it's a good idea, because this endpoint is a part of `/panel` route. 

Now it looks like this: `GET /panel/template`. It'd look like `GET /panel/panel-template`  with your suggestion.

server/src/index.ts
```diff
@@ -57,6 +58,7 @@ async function init() {
   rootRouter.use('/segments', segmentsRouter.routes(), segmentsRouter.allowedMethods());
   rootRouter.use('/query', dataRouter.routes(), dataRouter.allowedMethods());
   rootRouter.use('/detections', detectionsRouter.routes(), detectionsRouter.allowedMethods());
+  rootRouter.use('/panel', panelRouter.routes(), panelRouter.allowedMethods());
```
corpglory-dev: above we used the plural for names. should we do the same here?

server/src/services/export_service.ts
```diff
@@ -0,0 +1,29 @@
+import * as AnalyticUnit from '../models/analytic_units';
+import * as AnalyticUnitCache from '../models/analytic_unit_cache_model';
+import * as DetectionSpan from '../models/detection_model';
+import * as Segment from '../models/segment_model';
+
+
+export async function exportPanel(panelId: string): Promise<{
```
corpglory-dev: Is it guaranteed that different organizations cannot have two identical panel ids?

rozetko: Interesting, I'll check it out, thanks

server/src/index.ts
rozetko: Makes sense

server/src/services/export_service.ts
rozetko: Oh shi~, you're right: different organizations can have dashboards with the same dashboard id. We should support it.

rozetko: Created issue: https://github.com/hastic/hastic-server/issues/834

jonyrock-back:
what is "analytic unit template" and why did you make it?

rozetko:
@jonyrock-back "analytic unit template" is analytic unit config with some fields changed to variables, e.g.:
- panel ID
- grafana URL
- datasource URL

It's made for using the same analytic unit in different Hastic instances or dashboards

PR #843, opened 2020-02-13
Refactor get_bounds_for_segment method

VargBurz:
This PR refactors `get_bounds_for_segment` Anomaly detector's method.

## Changes:
 - use [`operator` lib](https://docs.python.org/3/library/operator.html) to remove duplicated code.
 - `get_bounds_for_segment` -> `get_segment_bound`: 
    - new required `bound: Bound` argument
    - returns `upper` or `lower` bound instead of both bounds depending on the passed `bound`

### DIFF
--- a/analytics/analytics/detectors/anomaly_detector.py
+++ b/analytics/analytics/detectors/anomaly_detector.py
@@ -4,6 +4,7 @@ import numpy as np
 import pandas as pd
 import math
 from typing import Optional, Union, List, Tuple, Generator
+import operator
 
 from analytic_types import AnalyticUnitId, ModelCache
 from analytic_types.detector_typing import DetectionResult, ProcessingResult, Bound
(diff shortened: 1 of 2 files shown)


analytics/analytics/detectors/anomaly_detector.py
```diff
@@ -195,54 +196,34 @@ def add_season_to_data(self, data: pd.Series, segment: pd.Series, offset: int, s
...
 
         return data[:len_smoothed_data]
 
-    def get_bounds_for_segment(self, segment: pd.Series) -> Tuple[pd.Series, pd.Series]:
+    def get_bound_for_segment(self, segment: pd.Series, bound: Bound) -> pd.Series:
         '''
-        segment is divided by the median to determine its top and bottom parts
-        parts are smoothed and raised so the segment is between them
+        segment is divided by the median to determine its top or bottom part
+        part is smoothed and raised above the segment or down under the segment
```
rozetko:
```suggestion
        the part is smoothed and raised above the segment or put down below the segment
```

Issue #842, opened 2020-02-12
Serialized cache / segment / detection span types


rozetko:
We should have "serialized" types for cache / segment / detection span models, like `SerializedAnalyticUnit`

jonyrock:
just for protocol, we don't need it cuz we can serialize to json

PR #844, opened 2020-02-14
Types for serialized cache / segment and detection span

VargBurz:
Closes https://github.com/hastic/hastic-server/issues/842
This PR adds types for serialized cache / segment / detection span


server/src/models/types.ts
```diff
@@ -0,0 +1,72 @@
+export type SerializedSegment = {
```
rozetko: Let's move these types into the corresponding models, e.g:
- `SerializedSegment` -> `models/segment`
- `SerializedCache` -> `models/analytic_unit_cache`
- etc.

PR #861, opened 2020-02-19
Panel export: change IDs #841

### LINKED ISSUE #841: Panel export: change IDs
Exported panel can't be imported within the same Grafana instance, there will be errors because of IDs collisions.

We should replace all IDs with placeholders during the exporting process.

rozetko:
Closes #841 

This PR changes panel importing process:  when panel is being imported, IDs of analytic units / caches / segments / detection spans are changed to new ones

## Disclaimer
Abbreviations in this PR:
- AU -> analytic unit
- AUC -> analytic unit cache
- S -> segment
- DS -> detection span

## New import process:
- remove IDs from AU / AUC / S / DS while exporting
- import AU, database generates new IDs for them
- import AUC / S / DS with new analytic unit IDs


rozetko:
@jonyrock-back 
Please, review it

server/src/services/grafana_service.ts
```diff
@@ -29,13 +31,51 @@ export async function importPanel(
   panelTemplate: GrafanaPanelTemplate,
   variables: GrafanaTemplateVariables
 ): Promise<void> {
+  const oldAnalyticUnitIds = panelTemplate.analyticUnits.map(analyticUnit => analyticUnit._id);
+
   panelTemplate.analyticUnits.forEach(analyticUnit => {
+    analyticUnit._id = undefined;
```
jonyrock-back: do you need to explicity set `analyticUnit._id = undefined`? Maybe you need `delete analyticUnit._id` here?

jonyrock-back: I personally think that you should warn if you find id in analytic unit, it not suppose to be here, right?
These id's should be removed on export, no?

server/src/services/grafana_service.ts
```diff
@@ -29,13 +31,51 @@ export async function importPanel(
...
-  await AnalyticUnit.insertMany(panelTemplate.analyticUnits);
-  await AnalyticUnitCache.insertMany(panelTemplate.caches);
-  await Segment.insertMany(panelTemplate.segments);
-  await DetectionSpan.insertMany(panelTemplate.detectionSpans);
+
+  const newAnalyticUnitIds = await AnalyticUnit.insertMany(panelTemplate.analyticUnits);
+
+  if(newAnalyticUnitIds.length !== oldAnalyticUnitIds.length) {
+    throw new Error(`
+      Something went wrong while inserting analytic units:
```
jonyrock-back: Please remove this line. This line would be relevant in outer catch block

server/src/services/grafana_service.ts
```diff
@@ -29,13 +31,51 @@ export async function importPanel(
...
+    })
+  );
+  const newDetectionSpans = panelTemplate.detectionSpans.map(
+    span => ({
+      ...span,
+      analyticUnitId: oldToNewAnalyticUnitIdsMapping.get(span.analyticUnitId)
+    })
+  );
+
+  await Promise.all([
```
jonyrock-back: what if one of these Promises breaks? You will partially insert data?
Yeah, I know it's a pain to make inserting right

server/src/services/grafana_service.ts
rozetko: Now it's not error: we don't remove IDs on export, but maybe this behavior should be changed

server/src/services/grafana_service.ts
rozetko: Yes, we should make some kind of transactions here

server/src/models/detection_model.ts
```diff
@@ -57,6 +57,14 @@ export class DetectionSpan {
     };
   }
 
+  public toTemplate(): any {
+    return {
+      ...this.toObject(),
+      _id: undefined,
```
jonyrock-back: same thing with `delete` here. Why do you set it to `undefined`?

rozetko: it's just more convenient here

PR #814, opened 2019-12-22
WIP: Websocket connectivity between server and analytics

jonyrock-back:
It's hard to support zmq on node.js side because of many reasons.
last version [zeromq.js](https://github.com/zeromq/zeromq.js/) supports only Node.js v.10.2+

all debug / production connectivity is removed and only websocket connection is available.

It fixes #753, #812 and #852 (closing port part)

### Server
* remove `zeromq` dependency from `server/package.json`
* add dependency to [websockets/ws](https://github.com/websockets/ws) and [@types/ws](https://www.npmjs.com/package/@types/ws) into `server/package.json`
* remove `zmqConectionString: ZMQ_CONNECTION_STRING,` from API info response (which has `/` url): not sure we should expose it anyway and it doesn't make sense since analytics connects to hastic server over http (same port as the API)
* remove from config `ZMQ_IPC_PATH`, `ZMQ_DEV_PORT`, `ZMQ_HOST` vars
* no `ZMQ_IPC_PATH` folder creation in `data_service.ts`
* no `npm rebuild zeromq` in node 6.14 build process
* server sends error if second instance of analytics tries to connect
* add [transform-object-rest-spread](https://www.npmjs.com/package/babel-plugin-transform-object-rest-spread) for transpiling the ws lib for node, [read more about that](https://github.com/babel/babel-loader/issues/170)

### Analytics
* add dependency to [aaugustin/websockets](https://github.com/aaugustin/websockets) in `analytics/requrements.txt`
* remove `ZMQ_DEV_PORT` and `ZMQ_CONNECTION_STRING` and replace it with `HASTIC_SERVER_URL`

## Notes
* we still use `pyzmq` for messaging between threads/actors, see `AsyncZmqThread`
* I encourage all supporters of this part to read comment in https://github.com/aaugustin/websockets/blob/master/src/websockets/client.py#L313 (and give a star to https://github.com/aaugustin/websockets)
* Give a star to https://github.com/websockets/ws
* Maybe we should send a webhock about bad analytic connect attempts
* there are optional deps to https://github.com/websockets/bufferutil and `utf-8-validate` which I don't know what to do with. 
* **the minumum supported version for the dev mode is 8.3.0** now. It's because of https://github.com/websockets/ws/issues/1703 (we don't use `babel-loader` in the dev mode)

## Known issues
* reconnecting in analytics to server on drop increases the number of ThreadPoolExecutor

PR #881, opened 2020-03-13
Fix websockets

rozetko:
This PR fixes the case when connection to server in analytics drops:
https://github.com/hastic/hastic-server/pull/814#issuecomment-588322888
https://github.com/hastic/hastic-server/pull/814#issuecomment-594709568

// TODO: more description


analytics/analytics/services/server_service.py
```diff
@@ -94,19 +97,32 @@ def __aiter__(self):
                 asyncio.ensure_future(self.__handle_ping())
             else:
                 asyncio.ensure_future(self._send_message_from_thread(received_string))
+
+    async def __reconnect(self):
+        if not self.__reconnecting:
+            self.__reconnecting = True
```
jonyrock-back: try to make a task which exists only in one instanse and try to await it here if it exist...or make a new task (which will be the only task)

jonyrock:
It's very important to explain what you found interesting during your research

jonyrock:
@rozetko please finish this work in two days

rozetko:
Moved to https://github.com/hastic/analytics/pull/17

PR #885, opened 2020-04-10
Fix analytic model tester tool

VargBurz:
This PR makes `analytic_model_tester.py` tool work with current pattern models.

### Motivation:
`analytic_model_tester.py` tool was added a year ago and has never been updated.
The analytics code has changed, so the tool doesn't work.

## Changes:
 - analytics testing begins with creating instance of `Detector` instead of `Model`
 - rename `Segment` class to `TesterSegment`
 -  `ModelData` class: required segments with `{ "from": int, "to": int }` type

### DIFF
--- a/analytics/tools/analytic_model_tester.py
+++ b/analytics/tools/analytic_model_tester.py
@@ -6,18 +6,19 @@ sys.path.extend([ANALYTICS_PATH, TESTS_PATH])
 import pandas as pd
 import numpy as np
 import utils
-import models
 import test_dataset
+from analytic_types.segment import Segment
+from detectors import pattern_detector, threshold_detector, anomaly_detector
 
 # TODO: get_dataset
 # TODO: get_segment
 PEAK_DATASETS = []
 # dataset with 3 peaks
-TEST_DATA = test_dataset.create_dataframe([0, 3, 5, 7, 5, 3, 0, 0, 1, 0, 1, 4, 6, 8, 6, 4, 1, 0, 0, 0, 1, 0, 3, 5, 7, 5, 3, 0, 1, 1])
-POSITIVE_SEGMENTS = [(1523889000000, 1523889000006), (1523889000021, 1523889000027)]
-NEGATIVE_SEGMENTS = [(1523889000009, 1523889000017)]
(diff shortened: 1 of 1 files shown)


analytics/tools/analytic_model_tester.py
```diff
@@ -91,9 +92,12 @@ def main(model_type: str) -> None:
         for data in PEAK_DATASETS:
             dataset = data.frame
             segments = data.get_segments_for_detection(1, 0)
-            model = models.PeakModel()
-            cache = model.fit(dataset, segments, 'test', {})
-            detect_result = model.detect(dataset, 'test', cache)
+            segments = [Segment.from_json(segment) for segment in segments]
+            detector = pattern_detector.PatternDetector('PEAK', 'test_id')
+            cache = detector.train(dataset, segments, {})
```
rozetko, line 97: should it be called `cache`?
maybe something like `training_result`?

analytics/tools/analytic_model_tester.py
```diff
@@ -6,18 +6,19 @@
...
 
 # TODO: get_dataset
 # TODO: get_segment
 PEAK_DATASETS = []
 # dataset with 3 peaks
-TEST_DATA = test_dataset.create_dataframe([0, 3, 5, 7, 5, 3, 0, 0, 1, 0, 1, 4, 6, 8, 6, 4, 1, 0, 0, 0, 1, 0, 3, 5, 7, 5, 3, 0, 1, 1])
-POSITIVE_SEGMENTS = [(1523889000000, 1523889000006), (1523889000021, 1523889000027)]
-NEGATIVE_SEGMENTS = [(1523889000009, 1523889000017)]
+TEST_DATA = test_dataset.create_dataframe([0, 0, 3, 5, 7, 5, 3, 0, 0, 1, 0, 1, 4, 6, 8, 6, 4, 1, 0, 0, 0, 1, 0, 3, 5, 7, 5, 3, 0, 1, 1])
+POSITIVE_SEGMENTS = [{'from': 1523889000001, 'to': 1523889000007}, {'from': 1523889000022, 'to': 1523889000028}]
```
rozetko, line 18:
```suggestion
# TODO: more convenient way to specify labeled segments
POSITIVE_SEGMENTS = [{'from': 1523889000001, 'to': 1523889000007}, {'from': 1523889000022, 'to': 1523889000028}]
```

PR #886, opened 2020-04-10
Use tasks in analytic model tester tool

VargBurz:
This PR updates `analytic_model_tester.py` tool: get info for learning and detection from tasks, same as analytics gets from server.

### Motivation:
We want to get info after all analytics process, not only after model process

## Changes:
 - Use `AnalyticUnitManager` instance instead of `Detector`
 - `ModelData` class:
     - new `get_task` method: create task for learning or detect
     - new `get_formated_segments`, `get_positive_segments`, `get_negative_segments` methods for simplify segments control
  - use `asyncio`: `main()` is async function now
  - `main()`: send learning and detect tasks to `AnalyticUnitManager` and await its results, get metrics from detection result.

### DIFF
--- a/analytics/tools/analytic_model_tester.py
+++ b/analytics/tools/analytic_model_tester.py
@@ -5,19 +5,22 @@ sys.path.extend([ANALYTICS_PATH, TESTS_PATH])
 
 import pandas as pd
 import numpy as np
-import utils
-import test_dataset
-from analytic_types.segment import Segment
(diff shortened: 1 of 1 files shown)


jonyrock:
@VargBurz let's close it in two days or remove this PR
@rozetko please check

VargBurz:
close and move this PR to [hastic-analytics](https://github.com/hastic/analytics/pull/16)

Issue #899, opened 2020-05-27
unfriendly crash with missing HASTIC_API_KEY


jonyrock:
why do we crash like this:
```
            throw new Error(`Please configure ${field}`);
            ^

Error: Please configure HASTIC_API_KEY
    at getConfigField (/home/corp/temp_projects/hastic/hastic-server/server/dist/server-dev.js:164:19)
    at Object../src/config.ts (/home/corp/temp_projects/hastic/hastic-server/server/dist/server-dev.js:125:26)
    at __webpack_require__ (/home/corp/temp_projects/hastic/hastic-server/server/dist/server-dev.js:20:30)
    at Object../src/services/data_layer/index.ts (/home/corp/temp_projects/hastic/hastic-server/server/dist/server-dev.js:3185:18)
    at __webpack_require__ (/home/corp/temp_projects/hastic/hastic-server/server/dist/server-dev.js:20:30)
    at Object../src/services/data_service/index.ts (/home/corp/temp_projects/hastic/hastic-server/server/dist/server-dev.js:3895:22)
    at __webpack_require__ (/home/corp/temp_projects/hastic/hastic-server/server/dist/server-dev.js:20:30)
    at Object../src/models/analytic_unit_cache_model.ts (/home/corp/temp_projects/hastic/hastic-server/server/dist/server-dev.js:994:24)
    at __webpack_require__ (/home/corp/temp_projects/hastic/hastic-server/server/dist/server-dev.js:20:30)
    at Object../src/controllers/analytics_controller.ts (/home/corp/temp_projects/hastic/hastic-server/server/dist/server-dev.js:268:27)
hastic-server crashed
```

when we can just show an error and exit without crashing

PR #901, opened 2020-05-27
Unfriendly crash with missing hastic api key #899

jonyrock:
fixes https://github.com/hastic/hastic-server/issues/899

#### `getConfigField` to `getConfigFieldAndPrintOrExit`
* I added "print" part to the name this function actually prints something
* I decided to exit from the main process (I know, it's bad to do it in a function) because we 1) don't catch these exceptions 2) wrap all constants in `config.ts` into `try` is pain

We will refactor this part when we get true services with async init, so please let's keep this workaround.

### DIFF
--- a/server/src/config.ts
+++ b/server/src/config.ts
@@ -27,12 +27,12 @@ export type DBConfig = {
 
 export const ANALYTICS_PATH = path.join(__dirname, '../../analytics');
 
-export const HASTIC_DB_IN_MEMORY = getConfigField('HASTIC_IN_MEMORY_PERSISTANCE', false);
+export const HASTIC_DB_IN_MEMORY = getConfigFieldAndPrintOrExit('HASTIC_IN_MEMORY_PERSISTANCE', false);
 // TODO: enum for DB types
-export const HASTIC_DB_CONNECTION_TYPE = getConfigField('HASTIC_DB_CONNECTION_TYPE', 'nedb', ['nedb', 'mongodb']);
+export const HASTIC_DB_CONNECTION_TYPE = getConfigFieldAndPrintOrExit('HASTIC_DB_CONNECTION_TYPE', 'nedb', ['nedb', 'mongodb']);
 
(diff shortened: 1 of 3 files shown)


server/src/config.ts
```diff
@@ -77,11 +81,14 @@ export const INSIDE_DOCKER = process.env.INSIDE_DOCKER !== undefined;
...
 // TODO: maybe rename it to "HASTIC_SERVER_ANALYTICS_URL"
-export const HASTIC_SERVER_URL = getConfigField('HASTIC_SERVER_URL', 'ws://localhost:8002');
-export const HASTIC_INSTANCE_NAME = getConfigField('HASTIC_INSTANCE_NAME', os.hostname());
+export const HASTIC_SERVER_URL = getConfigFieldAndPrintOrExit('HASTIC_SERVER_URL', 'ws://localhost:8002');
+export const HASTIC_INSTANCE_NAME = getConfigFieldAndPrintOrExit('HASTIC_INSTANCE_NAME', os.hostname());
 
 
-function getConfigField(field: string, defaultVal?: any, allowedVals?: any[]) {
+/**
+ * You get an value or exit from the main process
```
rozetko, line 89:
```suggestion
 * You get a value or exit from the main process
```

server/src/config.ts
```diff
@@ -96,13 +103,16 @@ function getConfigField(field: string, defaultVal?: any, allowedVals?: any[]) {
...
+      console.log(`Please configure ${field}`);
+      exit(EXIT_CODE_MISSING_FIELD);
     }
+
     val = defaultVal;
   }
 
   if(allowedVals !== undefined && !_.includes(allowedVals, val)) {
-    throw new Error(`${field} value must be in ${allowedVals}, got ${val}`);
+    console.log(`${field} value must be in ${allowedVals}, got ${val}`);
```
rozetko, line 114:
```suggestion
    console.log(`${field} value must be one of: ${allowedVals}, got ${val}`);
```

PR #940, opened 2021-01-18
Fix anomaly detector webhooks

rozetko:
Now we don't send an alert if at least 1 segment was merged while inserting. This PR fixes the behavior.

Changes:
- while inserting segments into the DB, check whether there are any new (non-merged) segments inserted, then send an alert

### DIFF
--- a/server/src/controllers/analytics_controller.ts
+++ b/server/src/controllers/analytics_controller.ts
@@ -73,7 +73,7 @@ export async function onDetect(detectionResult: DetectionResult): Promise<Segmen
     AnalyticUnit.setDetectionTime(id, payload.lastDetectionTime),
   ]);
 
-  if(insertionResult.removedIds.length === insertionResult.addedIds.length) {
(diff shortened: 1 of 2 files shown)


server/src/controllers/analytics_controller.ts
```diff
@@ -72,8 +72,11 @@ export async function onDetect(detectionResult: DetectionResult): Promise<Segmen
     AnalyticUnitCache.setData(id, payload.cache),
     AnalyticUnit.setDetectionTime(id, payload.lastDetectionTime),
   ]);
-  // removedIds.length > 0 means that there was at least 1 merge
-  if(insertionResult.removedIds.length > 0) {
+
+  if(insertionResult.removedIds.length === insertionResult.addedIds.length) {
+    if (insertionResult.removedIds.length > 0) {
```
rozetko, line 77:
```suggestion
    if(insertionResult.removedIds.length > 0) {
```

VargBurz (approved):

